PR #14, opened 2017-10-24
 Add -c, --current option to get current context

JulienBreux:
Hi Ahmet!

To begin, thank you very much for this a-ma-zing and useful tools.

This is my tiny contribution ;)

Have a good day! 🍻

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ USAGE:
   kubectx <NAME>            : switch to context
   kubectx -                 : switch to the previous context
   kubectx <NEW_NAME>=<NAME> : create alias for context
+  kubectx -c,--current      : get current context
   kubectx -h,--help         : show this message
 ```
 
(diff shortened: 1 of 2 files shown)


JulienBreux:
I signed it!

ahmetb:
@JulienBreux thanks for your contribution, may I ask what's your use case? When you run `kubectx` no args, we already highlight the current context.

ahmetb:
@JulienBreux wouldn't running `kubectl config view -o=jsonpath='{.current-context}'` directly be more effective in your use case? My suspicion is that you're trying to use kubectx for scripting, as if it's a solid idempotent UNIX tool. Whereas in reality it doesn't meet those standards and is meant to be used interactively.

JulienBreux:
@ahmetb I know, but it's very sad day to me 😂 
In fact, you'r right !
Because, it's very smart in interactive way to havs `kubectx -c` to get only the current env (not list with colors).

Be that as it may, thank you so mutch for your work @ahmetb !

ahmetb:
Hey no problem. All I'm trying to do is to make sure we're optimizing for a prime use case and not adding features we absolutely don't need. Thanks for your idea.

ahmetb:

JulienBreux:
@ahmetb ok :) Thx

PR #18, opened 2017-12-04
add a nocolor mode

rsalmond:
I wanted to be able to do this:

```
for cluster in `kubectx`; do kubectx $cluster && kubectl get pods; done
```

Without getting an error like this:

```
error: no context exists with the name: "\x1b[40m\x1b[33m<mycluster>\x1b(B\x1b[m".
```

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -36,6 +36,7 @@ usage() {
   cat <<"EOF"
 USAGE:
   kubectx                   : list the contexts
+  kubectx -nc               : list the contexts without colors
   kubectx <NAME>            : switch to context <NAME>
   kubectx -                 : switch to the previous context
   kubectx <NEW_NAME>=<NAME> : rename context <NAME> to <NEW_NAME>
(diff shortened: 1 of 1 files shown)


rsalmond:
I signed it!

rsalmond:
Srsly tho I signed the thing!

ahmetb:
You should not use `kubectx` for scripting.

Try:

    kubectl config get-contexts -o=name

which does exactly what you want.

rsalmond:
Fair enough @ahmetb. For scripting I agree this is not the tool for the job, for one off one liners in my shell it is quite handy though.

pswenson:
Related:  anyone know a good way to remove color codes on scripts?  I have run into this issue with other tools that I try to script and never found a reliable way to remove them.

ahmetb:
@rsalmond kubectl is already installed and brief enough , please use that 😊

@pswenson try: `command | cat`

PR #22, opened 2018-01-10
Don't use color in output if piping is used

toredash:
Suggest code change that will enable use-cases like this:
`for env in $(kubectx | grep staging); do kubectx $env && kubectl scale deployments nginx --replicas=10; done`


toredash:
I signed it!

ahmetb:
We got this exact same PR before. Unfortunately we don't want scripting around `kubectx`.

You should use `kubectl config get-contexts -o name` to list contexts in your script.

PR #27, opened 2018-02-21
Only print colours when connected to a terminal

gechr:
This is useful when using `kubectx` or `kubens` as part of a pipeline or when redirecting to a file and outputting colours' escape sequences is undesirable, i.e.
```
$ kubectx | less
$ kubens | grep "<ctx>"
$ kubectx > contexts.txt
$ kubens > namespaces.txt
```

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -47,9 +47,13 @@ EOF
 list_contexts() {
   set -u pipefail
   local cur="$(current_context)"
-  local yellow=$(tput setaf 3)
-  local darkbg=$(tput setab 0)
-  local normal=$(tput sgr0)
+
+  local yellow darkbg normal
+  if [[ -t 1 ]]; then
+    yellow=$(tput setaf 3)
+    darkbg=$(tput setab 0)
+    normal=$(tput sgr0)
+  fi
 
   for c in $(get_contexts); do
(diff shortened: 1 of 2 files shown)


gechr:
I signed it!

gechr:
Ergh, maybe I should search closed PRs in in future...
* https://github.com/ahmetb/kubectx/pull/18
* https://github.com/ahmetb/kubectx/pull/22

ahmetb:
😃 thanks! 

You can achieve these stuff with kubectl directly. Just peek at kubectx/kubens source code.

gechr:
I very much appreciate that, and that's how I currently am doing/will continue to do in scripts regardless of whether or not `kubectx` outputs escape sequences when not connected to a terminal.

Relying on a shell script that's not in my control for my personal scripts is not my intention here, I'd always opt to use `kubectl` directly in this case. My use case is quick one-liners, as a matter of convenience (e.g. `for c in $(kctx); echo $c; kubectl --context=$c get pods; fi`).

ahmetb:
This should be sufficient for your snippet

    for c in $(kubectl config get-contexts -o=name);do echo $c; kubectl --context=$c get pods; done

PR #28, opened 2018-02-21
Get current context with -c

seslattery:
This is a simple addition to add a -c flag to get the current context.  I've found it very useful, especially in shell prompts.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -38,6 +38,7 @@ USAGE:
   kubectx                   : list the contexts
   kubectx <NAME>            : switch to context <NAME>
   kubectx -                 : switch to the previous context
+  kubectx -c                : show current context
   kubectx <NEW_NAME>=<NAME> : rename context <NAME> to <NEW_NAME>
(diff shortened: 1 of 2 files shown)


seslattery:
I signed it!

ahmetb:
`kubectx` shows the current context already in a different color. Why do you need this?

What you're looking for is `kubectl config current-context`; not a new kubectx knob.

ahmetb:
Sorry I misunderstood. Your PR title is inaccurate: you're trying to show the current namespace.

You can also do that by running `kubens`, which shows the current namespace in a different color.

If you're trying to use "kubens -c" for scripting, do not do that and copy the logic in current_namespace function which is:

     kubectl config view -o=jsonpath="{.contexts[?(@.name==\"$(kubectl config current-context)\")].context.namespace}"

if empty, it means `default` namespace, otherwise it will return the ns of the current context.

ahmetb:

PR #32, opened 2018-03-10
Example installation steps on Linux

shiroyasha:
Example installation steps for Linux make it easier to install this package faster. 

I'm not 100% sure that the `/opt` directory is the best place to put it, but I wanted to avoid injecting `utils.bash` into `/usr/local/bin` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,14 @@ them to any POSIX environment that has Bash installed.
 - Make `kubectx` and `kubens` executable (`chmod +x ...`)
 - Figure out how to install bash/zsh/fish [completion scripts](completion/).
 
+Example installation steps:
+
+``` bash
+sudo git clone https://github.com/ahmetb/kubectx /opt/kubectx
+sudo ln -s /opt/kubectx/kubectx /usr/local/bin/kubectx
+sudo ln -s /opt/kubectx/kubens /usr/local/bin/kubens
+```
+
 -----
 
 ####  Users


ahmetb:
Thanks for this PR, I’m not sure about Linux convention on this either.
@jonmosco had some ideas about Linux packaging, this might be a good time to share those. 

I’m thinking about getting rid of utils.bash and just duplicating its contents in each executable. Currently it causes more hassle than benefit for Linux.

jonmosco:
I have created packages for Debian, as well as a snap that that I am currently testing.  The only issue so far is with the utils.bash file due to the install paths that are used when packaging.  If those functions could be included in the executables like you mentioned, that would be great.

As far as RPMs are concerned, getting this into EPEL should not be to hard.

`distribution`
For the Debian packages, I am not sure what 3rd party repositories these can be upload to.  Any suggestions for that would be very helpful.

ahmetb:
@jonmosco can you send a PR incorporating `utils.bash` into both executables?

jonmosco:
Sure thing, on its way.

jonmosco:
Did you get a chance to create a new release?  Once that is ready, I can start the process to have a package submitted to the Debian project (can sometimes be lengthy).

ahmetb:
I'll merge this and tag a 0.4.1, just because we removed utils.bash. Someone needs to go update the Homebrew formula, too. See #34.

PR #33, opened 2018-03-15
Incorporate utils.bash into each executable to help simplify packaging for Linux and other platforms.  This adds to PR #32

jonmosco:
* Incorporate functions from utils.bash into kubectx and kubens
* Remove utils.bash (no longer needed)
* Updated README.md

Both executables locally tested and are in proper working order.


kubectx
```diff
@@ -44,6 +35,18 @@ EOF
...
 
+current_context() {
+  kubectl config view -o=jsonpath='{.current-context}'
+}
+
+get_contexts() {
+  kubectl config get-contexts -o=name | sort -n
+}
+
+get_namespaces() {
```
ahmetb: function not used in this file

kubens
```diff
@@ -53,6 +44,18 @@ current_namespace() {
   fi
 }
 
+current_context() {
+  kubectl config view -o=jsonpath='{.current-context}'
+}
+
+get_contexts() {
```
ahmetb: function not used in this file

ahmetb:
Thanks, now I suspect the Homebrew formula needs updating: https://github.com/Homebrew/homebrew-core/blob/8d530a70f8fcf02129d924090d73efbe12ff438d/Formula/kubectx.rb#L17

Once I tag a release, it will break the formula. Let me think how we can coordinate this.

PR #37, opened 2018-04-04
Allow for '/' in context name

jvassev:
Also, fix `kubens -` which fails to save previous namespace
after 6610d70

### DIFF
--- a/kubens
+++ b/kubens
@@ -53,8 +53,12 @@ get_namespaces() {
   kubectl get namespaces -o=jsonpath='{range .items[*].metadata.name}{@}{"\n"}{end}'
 }
 
+escape_context_name() {
+  echo "${1//\//-}"
+}
+
 namespace_file() {
-  local ctx="${1}"
+  local ctx="$(escape_context_name "${1}")"
   echo "${KUBENS_DIR}/${ctx}"
 }
 
@@ -62,6 +66,7 @@ read_namespace() {
   local f
   f="$(namespace_file "${1}")"
   [[ -f "${f}" ]] && cat "${f}"
+  return 0
 }
 
 save_namespace() {


jvassev:
I signed it

kubens
```diff
@@ -54,14 +54,15 @@ get_namespaces() {
...
-  local ctx="${1}"
+  local ctx="${1//\//-}"
   echo "${KUBENS_DIR}/${ctx}"
 }
 
 read_namespace() {
   local f
   f="$(namespace_file "${1}")"
   [[ -f "${f}" ]] && cat "${f}"
+  return 0
```
ahmetb: I'm trying to understand why this exists. Can you explain?

kubens
```diff
@@ -54,14 +54,15 @@ get_namespaces() {
 }
 
 namespace_file() {
-  local ctx="${1}"
+  local ctx="${1//\//-}"
```
ahmetb: do you mind creating a 

```sh
escape_context_name() {
    echo "${1//\//-}"
}
```

and using it here?

kubens
jvassev: As you said in  6610d70, the https://github.com/koalaman/shellcheck/wiki/SC2155 prevents masking the return value. In case $f does not exists the return value is 1, and since you're using `set -eou` this aborts the script.

A non-existent file in this case is just fine, so I again mask the return value

kubens
jvassev: sure, added it.

jvassev:
I am renaming the contexts generated by a tool :) There are just too-many-dashes-in-the-name. I needed a delimiter other than - or _ and one that implies hierarchy. Slashes won the contest :)

kubens
ahmetb: I see it now when I executed it myself with `DEBUG=1 ./kubens`. Thanks so much for noticing this.

I guess this works, because in both places `read_namespace` is called, empty output will do the job.

ahmetb:
@jvassev Thank you for your patch and debugging the subtle issue with the return value.

If you're interested in development of the project, please hit "Watch" so you can participate in issue discussions and feature proposals.

PR #46, opened 2018-05-18
Allow deleting multiple contexts at once

### LINKED ISSUE #39: kubectx -d: support multiple arguments
I wish to see `kubectx -d` used to delete multiple commands at once. Ideally I should be able to say:

    kubectx -d ctx1 ctx2

to delete multiple contexts.

uesteibar:
fixes #39

Example: `kubectx -d ctx1 ctx2`

This can be useful when doing things like `kubectx -d $(kubectx | grep test)`

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -26,14 +26,14 @@ KUBECTX="${HOME}/.kube/kubectx"
 usage() {
   cat <<"EOF"
 USAGE:
-  kubectx                   : list the contexts
-  kubectx <NAME>            : switch to context <NAME>
-  kubectx -                 : switch to the previous context
-  kubectx <NEW_NAME>=<NAME> : rename context <NAME> to <NEW_NAME>
-  kubectx <NEW_NAME>=.      : rename current-context to <NEW_NAME>
-  kubectx -d <NAME>         : delete context <NAME> ('.' for current-context)
-                              (this command won't delete the user/cluster entry
-                              that is used by the context)
+  kubectx                       : list the contexts
+  kubectx <NAME>                : switch to context <NAME>
+  kubectx -                     : switch to the previous context
+  kubectx <NEW_NAME>=<NAME>     : rename context <NAME> to <NEW_NAME>
+  kubectx <NEW_NAME>=.          : rename current-context to <NEW_NAME>
+  kubectx -d <NAME> [<NAME...>] : delete context <NAME> ('.' for current-context)
+                                  (this command won't delete the user/cluster entry
+                                  that is used by the context)
 
   kubectx -h,--help         : show this message
 EOF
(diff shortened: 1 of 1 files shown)


uesteibar:
I signed it!

ahmetb:
Thanks so much for this PR. I'll review now.

kubectx
```diff
@@ -32,6 +32,7 @@ USAGE:
   kubectx <NEW_NAME>=<NAME> : rename context <NAME> to <NEW_NAME>
   kubectx <NEW_NAME>=.      : rename current-context to <NEW_NAME>
   kubectx -d <NAME>         : delete context <NAME> ('.' for current-context)
+                              accepts multiple context names separated by a space
```
ahmetb: I think the right notation is to do:

     kubectx -d <NAME> [<NAME...>]

then you don't need this line.

kubectx
```diff
@@ -148,6 +149,13 @@ rename_context() {
   kubectl config rename-context "${old_name}" "${new_name}"
 }
 
+delete_contexts() {
+  IFS=' ' read -ra CTXS <<< "${1}"
```
ahmetb: I don't think you need this. 

See:


```bash
#!/bin/bash

f() {
    for v in "${@}"; do
      echo "--> ${v}"
    done
}

f "${@:2}"
```

this still iterates correctly.

ahmetb:
One of my hopes is to get the bash/zsh completion to work with this command, but so far I have not been able to find someone who knows about this topic to fix that.

I think having tab completion would totally be useful with this command but I won't require you to fix the scripts.

kubectx
uesteibar: Actually, I did some tests here and in this function `${1}` equals `"ctx1 ctx2"` (a string, not an array). It seems that `"${@:2}` mashes the arguments into a string, so I do need to split it afterwards.

After some research I found [this](https://unix.stackexchange.com/a/92985), which seems to explain what's happening here.

> When in scalar context the list of elements is joined with space characters with `bash` and `ksh93` and with the first character of `$IFS` (or nothing if `$IFS` is empty or space if it is unset) with `zsh`.

Would you be fine with keeping the current solution as it is? I didn't manage to come up with a solution to overcome this issue.

ahmetb: This might be ok. I've been using `mapfile -t` to turn strings into arrays:

    mapfile -t my_arr < <(kubectl get pods -o=name)

This only works on bash 4.0+. So most macOS users still can't use this with /bin/bash.

`read` solution is slower, but works, so we can keep it.

uesteibar: That's nice, didn't know about this. Let's keep this one then :) the rest of the feedback is already addressed.

PR #48, opened 2018-05-23
Make the scripts portable

xnaveira:
Some unix like systems (including some linux distributions) put the bash interpreter in other locations than `/bin`. By using `env` the scripts work on those as well.


ahmetb:
Thanks!
Do you mind amending the commit with the right email address that has the CLA signed?

PR #49, opened 2018-05-28
Use * to indicate current context & current namespace

prabhu43:
For the people using [Flux](https://justgetflux.com/), It is difficult to identify the current context and current namespace which is highlighted in yellow colour.

Hence, to make it clear, added an asterisk(*) to indicate the same.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -60,7 +60,7 @@ list_contexts() {
 
   for c in $(get_contexts); do
   if [[ -t 1 && "${c}" = "${cur}" ]]; then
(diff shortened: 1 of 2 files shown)


ahmetb:
This won't work, as people use `$(kubectx)` and `$(kubens)` outputs in scripts etc. So we need to preserve the context name as is. We can look at changing the color to red etc.

prabhu43:
okay @ahmetb. What about the configuring the colour using env variable?

PR #51, opened 2018-05-30
Add command attribute to show current context

timusg:
Currently kubens display current namespace in the color formatted way in the namespace list, which is difficult to grep and find the current namespace.

the separate attribute to display current namespace can be helpful in finding it and then it can be easily utilized for tmux status or zsh prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ USAGE:
   kubens                    : list the namespaces
   kubens <NAME>             : change the active namespace
   kubens -                  : switch to the previous namespace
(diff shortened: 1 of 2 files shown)


timusg:
I signed it!

ahmetb:
Thanks, but same PR arrived many more times. You should use `kubectl config view current-context`. #14, #28.

You're not supposed to use kubectx/kubens with scripting. They're meant to be interactive convenience programs that just wrap kubectl.

ahmetb:

PR #55, opened 2018-06-07
Import kubeconfig using -i

jvassev:
Add `-i` which import another kubeconfig at the default location.

It will never overwrite an existing context with the same name but instead will rename the incoming context just before import. Also, it won't touch the provide file.
After import, the user can rename the context using new=old syntax or optionally delete an old (replacing it).

Bash completion works - filenames are completed after -i but I'm not an expert and someone more experienced with `compgen` needs to have a look :)

Example:
```bash
$ kubectx
foo
bar

# import /tmp/hello.yaml which contains context named bar
$ kubectx -i /tmp/hello.yaml
kubectx -i /tmp/hello
Importing kubeconfig from /tmp/hello:
  + hello.yaml-c731b0dc21b4-bar

# c731b0dc21b4 is the first 12 bytes of the sha256 sum of /tmp/hello.yaml
# if the same file is imported again no new contexts will be created
```

### DIFF
--- a/completion/kubectx.bash
+++ b/completion/kubectx.bash
@@ -1,8 +1,14 @@
 _kube_contexts()
 {
-  local curr_arg;
+  local curr_arg
   curr_arg=${COMP_WORDS[COMP_CWORD]}
-  COMPREPLY=( $(compgen -W "- $(kubectl config get-contexts --output='name')" -- $curr_arg ) );
+  if [[ "${COMP_WORDS[$(($COMP_CWORD - 1))]}" == "-i" ]]; then
+    compopt -o default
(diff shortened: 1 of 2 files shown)


ahmetb:
First of all, please propose a feature without implementing it. This will save you time if it gets rejected.

Second, if you want to use a different KUBECONFIG file, just set the KUBECONFIG in your environment.

If you want to merge kubeconfigs, run `KUBECONFIG=1.yaml:2.yaml kubectl config view --minify --flatten > out.yaml` .

kubectx should not be in the business of kubeconfig handling, as it calls kubectl under the covers, which knows about kubeconfig.

PR #64, opened 2018-07-18
Debian package installation information added

idvoretskyi:
As the package is currently available for Debian Buster and Sid, the brief installation information added.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,17 +111,20 @@ sudo ln -s /opt/kubectx/kubectx /usr/local/bin/kubectx
 sudo ln -s /opt/kubectx/kubens /usr/local/bin/kubens
 ```
 
-Also available as a Debian package (for [Debian Buster (testing)](https://packages.debian.org/buster/kubectx) and [Sid (unstable)](https://packages.debian.org/sid/kubectx)):
-
-``` bash
-sudo apt install kubectx
-```
-
 #### Arch Linux
 
 An unofficial [AUR package](https://aur.archlinux.org/packages/kubectx) `kubectx`
 is available. Install instructions can be found on the [Arch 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,13 @@ sudo git clone https://github.com/ahmetb/kubectx /opt/kubectx
 sudo ln -s /opt/kubectx/kubectx /usr/local/bin/kubectx
 sudo ln -s /opt/kubectx/kubens /usr/local/bin/kubens
 ```
+
```
ahmetb: Do you mind doing this:

```
### Debian/Ubuntu

....

### Other Linux

(the instructions above this line basically)
```

README.md
```diff
@@ -110,6 +110,13 @@ sudo git clone https://github.com/ahmetb/kubectx /opt/kubectx
...
+Also available as a Debian package (for [Debian Buster (testing)](https://packages.debian.org/buster/kubectx) and [Sid (unstable)](https://packages.debian.org/sid/kubectx)):
+
+``` bash
+sudo apt install kubectx
```
ahmetb: can you provide the commands to add one of the package repos above? it would be much better if folks can copy paste and it works (rather than trying to figure out how to add buster or sid).

README.md
idvoretskyi: I'd add "Other Linux" line after we'll have enough content for the distro sections, but added the explicit section mentioning Debian/Ubuntu.

ahmetb: I mostly said that because in terms of Desktop usage it probably goes like:

- Debian/Ubuntu
- Arch Linux 
- Other Linux 

(not sure about the last two but it feels like Arch should appear first, then other as a "catch all"). But this is fine as is.

Issue #68, opened 2018-08-06
zsh tab completion not working when 2+ contexts


ahmetb:
```
$ kubectx
docker-for-desktop
gke_ahmetb-samples-playground_australia-southeast1-a_eu
gke_ahmetb-samples-playground_us-central1-b_demo
gke_istio-next-2_us-central1-b_demo
gke_istio-next-3_us-central1-b_demo
gke_istio-next-4_us-central1-b_demo
gke_istio-next-4_us-east4-a_east-coast
gke_istio-next-5_us-central1-a_istio-service-mesh
gke_istio-next-5_us-central1-b_demo
gke_istio-next_us-central1-b_demo
gke_microservices-demo-app_us-central1-b_demo-app
```

but when I do `kubectx`<kbd>Tab</kbd><kbd>Tab</kbd> I get only two items:

```
$ kubectx -
-                   docker-for-desktop
```

which is basically the hardcoded `-` and the **first context** in the list.

When I edit the _kubectx script:

```
  9     _arguments "1: :((- \
 10         "${contexts[@]}" \
 11 |_______))"
```

to add a hardcoded name like:

```
  9     _arguments "1: :((- \
 10         "${contexts[@]}" \
 11         hardcoded_cluster
 12 |_______))"
```

tab completion shows everything correctly.

What's going on here?

czchen:
I guess it is caused by different separators (` ` for `-` and `\n` for others) used in `_arguments`. Just change `-` separator to `\n` can solve the problem. Like the following:

```
    _arguments "1: :($(kubectl config get-contexts --output='name')
-)"
```

ahmetb:
This is so weird and it only reproduces with the set of contexts listed above.

ahmetb:
I tried multiple things here, like deleting the new line between `-` and the list:
```
  8     _arguments "1: :((- $(kubectl config get-contexts --output='name') ))"
```

adding new line after list, with `\`:

```
  8     _arguments "1: :((- \
  9         $(kubectl config get-contexts --output='name') \
 10         ))"
```

without  `\`:

```
  8     _arguments "1: :((-
  9         $(kubectl config get-contexts --output='name')
 10         ))"
```

no luck, the problem still reproduces.

czchen:
Try the following one? It works on my machine.

```
    _arguments "1: :(- $(kubectl config get-contexts --output='name'))"
```

I guess the problem is not `\n`, but the additional parenthesis.

[...]

PR #77, opened 2018-08-23
Fix kubectx zsh tab completion

czchen:
Closed: #68

### DIFF
--- a/completion/kubectx.zsh
+++ b/completion/kubectx.zsh
@@ -5,8 +5,8 @@ PREV=""
 if [ -f "$KUBECTX" ]; then
     # show '-' only if there's a saved previous context
     local PREV=$(cat "${KUBECTX}")
-    _arguments "1: :((- \
-        $(kubectl config get-contexts --output='name')))"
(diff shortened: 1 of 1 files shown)


completion/kubectx.zsh
```diff
@@ -5,8 +5,8 @@ PREV=""
 if [ -f "$KUBECTX" ]; then
     # show '-' only if there's a saved previous context
     local PREV=$(cat "${KUBECTX}")
-    _arguments "1: :((- \
-        $(kubectl config get-contexts --output='name')))"
+    _arguments "1: :($(kubectl config get-contexts --output='name')
+-)"
```
ahmetb: we can't do this. `-` has to be the first option.

can you reproduce the problem described in the bug?

ahmetb:
This bug doesn't repro anymore. I'll close.

ahmetb:
The proposed fix works but it's moving `-` to the end, which is not something I wanna do. See my updated message on the original issue. If you can reproduce the bug locally and solve it with `-` at the beginning, I'd be happy to accept another PR.

PR #80, opened 2018-08-27
README: Install completion for zsh & Linux

schnatterer:
It's not exactly intuitive to use the code completion in zsh on Linux, so I added some hints to get started to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,8 @@ them to any POSIX environment that has Bash installed.
 - Figure out how to install bash/zsh/fish [completion scripts](completion/).  
   - For zsh:
    ```bash
-   sudo ln kubectx/completion/kubectx.zsh /usr/share/zsh/functions/Completion/Zsh/_kubectx.zsh
-   sudo ln kubectx/completion/kubens.zsh /usr/share/zsh/functions/Completion/Zsh/_kubens.zsh`
+   sudo ln -s kubectx/completion/kubectx.zsh /usr/share/zsh/functions/Completion/Zsh/_kubectx.zsh
+   sudo ln -s kubectx/completion/kubens.zsh /usr/share/zsh/functions/Completion/Zsh/_kubens.zsh`
    ```
 
 Example installation steps:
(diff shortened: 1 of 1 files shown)


schnatterer:
I signed it!

ahmetb:
I'm tempted to reject this patch, although thanks for sending it.

I think the path you provided simply won't work widely enough.

1. most popular zsh distribution (oh-my-zsh) doesn't install at this path.
2. this documents it only for zsh (leaving out more popular shell, "bash")
3. ln probably should be "ln -s" and/or it should use a full path. 

If you need completion yourself, try out the new fzf interactive selection support we added in v0.6 (see README).

schnatterer:
I've been using kubectx for some time now, but since I did't get the tab autocompletion to work right away, I used it without tab autocompletion for the time being. 

You tweet about fzf actually brought me back here and I'd thought I 'd give setting up autocompletion another go. I love fzf but I think it's just something else than tab completion.

This time I got it to work, altough it took me some time to figure out how to get it to work.
I was about to write down how to set it up for myself when I thought I might as well write it down for everyone else via this PR. 
This is acutally how I got it to work *with oh-my-zsh*.

Yes, your right it does not document bash and might not work for everybody. Still it contains more info than "Figure out how to install bash/zsh/fish" (e.g. the link to *underscore*`kubens.zsh`).

Maybe someone else get motivated to document bash and fish here.

Ah and, you were also right, I used a symlink, corrected it in a second commit. Feel free to squash.

ahmetb:
Thanks for updating the PR, do you mind exploring the actual paths on Linux/Mac (+with oh-my-zsh). I assume oh-my-zsh paths at least will be predictable. 

Or at least a note saying "these paths may not exist depending on your OS or zsh installation method, find the actual method".

schnatterer:
I did some more research and extended the docs.
Can't say if that'll work on Mac, though. It's in the Linux section of the README after all ;-)

ahmetb:
@schnatterer can you please check again if you signed the CLA? It doesn't show up in Google's CLA records (make sure you use the same email address and/or github username in your commit).

ahmetb:
Your email address looks like this, probably because you initiated the update from Github web interface:

    From: schnatterer <schnatterer@users.noreply.github.com>

Feel free to do:

    git clone <your fork>
    git checkout -b patch-1
    git commit --amend --author "Your Name <you@example.com>"
    git push -f

on your developer machine.

PR #82, opened 2018-08-29
Fzf toggle pr

oliverisaac:
I use FZF for other things but didn't want it enabled for this tool. This PR allows EU's to selectively disable FZF for both tools or just one tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ with fuzzy searching, you just need to [install
 
 ![kubectx interactive search with fzf](img/kubectx-interactive.gif)
 
+If you have `fzf` installed, but want to opt out of using this feature, set the environment variable `KUBECTX_IGNORE_FZF=1`.
+
+
 -----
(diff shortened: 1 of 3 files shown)


oliverisaac:
I signed it!

ahmetb:
ooh, this is exiciting!!!

I think we should just call it `KUBECTX_IGNORE_FZF` and use it in both kubens/kubectx (I don't see why people would disable it for one and keep it enabled for another). Can you make the change?

Also we should add to the readme something like:

> If you have `fzf` installed, but want to opt out of using this feature, set the environment variable `KUBECTX_IGNORE_FZF=1`.

oliverisaac:
Done and done. I definitely agree with the env var name as NO_FZF felt a little to generic.

ahmetb:
Just tested manually, looks good thanks. It'll ship in the next release.

PR #86, opened 2018-09-13
Added example of using bash completion

tc-turner:
Added example using bash completion.

Example clones into `~/.kubectx` for people who don't want to make modifications outside of `$HOME`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,7 @@ them to any POSIX environment that has Bash installed.
     export PATH=~/.kubectx:\$PATH
     . ~/.kubectx/*.bash
     FOE
+    . ~/.bashrc
     ```
   - For fish: Figure out how to install completion scripts and please document here
   
--- a/README.md
+++ b/README.md
@@ -114,7 +114,16 @@ them to any POSIX environment that has Bash installed.
     Note that the leading underscore seems to be a convention.  
     If not using oh-my-zsh, you could link to `/usr/share/zsh/functions/Completion` (might require sudo), depending on the `$fpath` of your zsh installation.  
     In case of error, calling `compaudit` might help.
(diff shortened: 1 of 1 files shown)


tc-turner:
I signed it!

ahmetb:
I'm not sure if sourcing the completion files directly is a good idea.

These are meant to be copied into some `bash_completion.d` directory (https://serverfault.com/questions/506612/standard-place-for-user-defined-bash-completion-d-scripts).

tc-turner:
I've update the bash example to use what is outlined in [scop/bash-completion](https://github.com/scop/bash-completion/blob/master/README.md#faq).  Using `~/.bash_completion` would require appending the content of the completion scripts to the end of that file and would be messy to update in the future.  


> **Q. I author/maintain package X and would like to maintain my own
>    completion code for this package. Where should I put it to be sure
>    that interactive bash shells will find it and source it?**
> 
> A. Install it in one of the directories pointed to by
>    bash-completion's `pkgconfig` file variables. There are two
>    alternatives:
> 
>    - The recommended directory is `completionsdir`, which you can get with
>    `pkg-config --variable=completionsdir bash-completion`. From this
>    directory, completions are loaded on-demand based on invoked commands' names,
>    so be sure to name your completion file accordingly, and to include
>    (for example) symbolic links in case the file provides completions
>    for more than one command.
>    - The other directory (which only present for backwards compatibility)
>      is `compatdir` (get it with
>    `pkg-config --variable=compatdir bash-completion`) from which files
>    are loaded when `bash_completion` is loaded.

I saw some references to `~/.bash_completion.d` but that path doesn't seem to be used by bash completion without extra configuration.

How does that look?

funky81:
thanks for this pull request. I can install this under centos 7

PR #93, opened 2018-10-26
Support XDG_CACHE_HOME environment variable

rafi:
By using the `XDG_CACHE_HOME` environment variable, people who set kubectl's config file elsewhere (using the `KUBECONFIG` environment variable) could now bypass the error that `~/.kube` directory doesn't exist, and also support the XDG standard :)

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -22,7 +22,7 @@ set -eou pipefail
 IFS=$'\n\t'
 
 SELF_CMD="$0"
-KUBECTX="${HOME}/.kube/kubectx"
+KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
 
 usage() {
   cat <<"EOF"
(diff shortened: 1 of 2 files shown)


rafi:
I signed it!

ahmetb:
@rafi thanks for your patch. Does kubectl actually respect `XDG_CACHE_HOME` today to locate your .kube/config file? If so, we can follow suit.

rafi:
@ahmetb No, it doesn't. But it supports the `KUBECONFIG` set a single or multiple configuration files, e.g. `export KUBECONFIG="$XDG_CONFIG_HOME/kube/config.yml"`

ahmetb:
I see we're using ~/.kube for caching stuff, so this will work. But you will still need to set KUBECONFIG to actually make things work.

I'll think a bit to see if this matches XDG_CACHE_HOME, XDG_CONFIG_HOME or XDG_DATA_HOME. But I think your patch is fine, I may merge it soon.

ahmetb:
Thanks a lot for your PR.

PR #96, opened 2018-11-01
Check for kubectl.exe on Windows

metcalfc:
On the various flavors of bash for Windows kubectl won't resolve as the binary is kubectl.exe.
Simple aliasing doesn't seem to work. So test for both otherwise fail with a not found error.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -41,11 +41,11 @@ EOF
 }
 
 current_context() {
-  kubectl config view -o=jsonpath='{.current-context}'
+  $KUBECTL config view -o=jsonpath='{.current-context}'
 }
 
 get_contexts() {
-  kubectl config get-contexts -o=name | sort -n
+  $KUBECTL config get-contexts -o=name | sort -n
 }
 
 list_contexts() {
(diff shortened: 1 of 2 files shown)


metcalfc:
I signed it!

ahmetb:
This looks good, let's hold off until I fully understand why bash isn't able to find  kubectl.exe. I'll probably move the logic outside main() (to the beginning of the script).

metcalfc:
Bash does find it if it is fully specified. Here is a scenario. 

Docker Desktop is installed on Windows 10 and `docker` and `kubectl` are exe binaries installed in some path on Windows. The user is choosing to run Bash in WSL and append the Windows path so that they don't have to do the hack where they also have to install `docker` and `kubectl` in the WSL side and point them back to the Windows side (via copying kube config or setting DOCKER_HOST). 

In a WSL bash shell, `kubectl.exe` is available `kubectl` isn't because the executable is in fact named`kubectl.exe` on the windows side. On the windows side, the cmd or PowerShell take care of not having to add the `.exe`. The way I hack this now is to hijack bash to make symbolic links so that something named `kubectl` is available in the path. The specifics of that hack are here: https://github.com/ahmetb/kubectx/issues/97#issuecomment-435277728

ahmetb:
Tested, looks good. Thanks a lot.

PR #62, opened 2018-07-11
add \? command to get the current namespace 

lindenle:
resolves #61

Issue #61, opened 2018-07-11
kubens command to display current namespace


lindenle:
It would be nice to have a command like `kubens current` that would return the current namespace. I would like this for instance to add the current namespace to my command prompt.

ahmetb:
I'm finally coming around to implement a feature to show current context/namespace. Please vote for it at https://github.com/ahmetb/kubectx/issues/127, I also would like to hear your opinions.

PR #43, opened 2018-04-23
adds an option to print the current context/namespace

wonko:
I needed an option to see the current namespace, and just leveraged what was already present in the tools. Did the effort for context too, for uniformity.

PR #100, opened 2018-12-12
 Add feature and flag to print current context / namespace

ewjmulder:
Sometimes we forget what the current context / namespace is and would like an easy way to just print it in the console. So I've added this simple feature to this great script. Furthermore, we use it to create a full listing of the current context / namespace / pods, etc.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -30,13 +30,14 @@ USAGE:
   kubectx                       : list the contexts
   kubectx <NAME>                : switch to context <NAME>
   kubectx -                     : switch to the previous context
+  kubectx -c,--current          : print the current context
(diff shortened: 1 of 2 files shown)


ahmetb:
See #28 #43 #51 #61 #62.

ewjmulder:
Oops, didn't see those same PR's.
But given the fact that several people requested this and took the effort of actually creating a PR for it, it seems like it's a desired feature and users will appreciate it. So maybe you can reconsider.

I will keep my local copy with the extra flag cause it's a useful feature for me. Printing all with a highlight is not workable, since we have dozens of namespaces.

Thanks for the nice script either way.

ahmetb:

PR #101, opened 2018-12-14
implement "-c NAME NEW_NAME" for copying contexts

joemiller:
This PR implements functionality to allow copying contexts. My use case for this is primarily for making it easier for folks to work with GKE clusters while still having the canonical GKE context name available.

For example, an engineering team may have a gke cluster named `gke_my-project_us-central1_cluster-01`. And this engineering team may have common tooling that expects a context of that name to be available. Also, engineers on this team want to be able to address this cluster as simply `cluster-01`. With the ability to copy a context to a new name the engineers on this team can make a friendly short name and not risk breaking any common tooling that expects the longer, default context name provided by GKE.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -35,6 +35,8 @@ USAGE:
   kubectx -d <NAME> [<NAME...>] : delete context <NAME> ('.' for current-context)
                                   (this command won't delete the user/cluster entry
                                   that is used by the context)
+  kubectx -c <NAME> <NEW_NAME>  : copy context <NAME> ('.' for current-context)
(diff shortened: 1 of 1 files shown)


ahmetb:
I don’t think this is a common use case. We have rename functionality for common case.

I recommend resorting to what kubectl config command provides for your task. Also please send a proposal first before implementing. 😊

joemiller:
@ahmetb Thanks. I tried to elaborate on the rationale here: https://github.com/ahmetb/kubectx/issues/102

Issue #98, opened 2018-11-30
Current cluster not highlighted in interactive mode


ahmetb:
I don't think this ever worked with fzf.

![image](https://user-images.githubusercontent.com/159209/49306242-8b9e4300-f486-11e8-8321-98e3c1d48ef5.png)

Issue #89, opened 2018-10-09
Must show colors for the current context in fzf mode


ahmetb:
When fzf installed, kubectx v0.6.1 is not showing any colors about current context. 

I think we need something like KUBECTX_FORCE_COLOR while invoking fzf. Need to investigate whether --ansi flag on fzf will be a problem or not.

PR #107, opened 2018-12-22
Show visual indication of the selected namespace and context

FabioAntunes:
When using `kubens` with `fzf` there's no visual indication of the selected namespace and context.
That's because `fzf` is considered a non-interactive tty, so there's no colour on the output.

This PR aims to fix that by creating a specific flag, that gets set every time we call kubens with fzf.

Fixes https://github.com/ahmetb/kubectx/issues/98 and https://github.com/ahmetb/kubectx/issues/89

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,8 +127,13 @@ them to any POSIX environment that has Bash installed.
     export PATH=~/.kubectx:\$PATH
     FOE
     ```
-  - For fish: Figure out how to install completion scripts and please document here
-  
+  - For fish:
+    ```fish
+    mkdir -p ~/.config/fish/completions
+    ln -s /opt/kubectx/completion/kubectx.fish ~/.config/fish/completions/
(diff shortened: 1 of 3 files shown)


FabioAntunes:
I signed it!

ahmetb:
I'm currently -1 for introducing yet another variable, some reasons:

- becomes yet another knob that takes effect only when fzf exists
- variable contains fzf (this was my bad, I have an existing variable like this) apparently other programs that do the same (e.g. `fzy`) exist too, so we need a generic name
- variable currently named KUBENS_ ideally it should be KUBECTX_


Simplest (but hacky) way I'm imagining we can solve this:

1. Force sending colors to fzf
2. fzf will show the colors
3. when user chooses an option, we'll get value with color sequences in it
4. strip off color sequences and use.

thoughts?

ahmetb:
Please review #109. The `if` logic there is a bit different than yours, I wanna make sure we get it right.

PR #108, opened 2018-12-22
test: enable travis-ci with bats

ahmetb:
- add .travis.yml.
- move bats fixtures to .bats extension, since it allows detection of test
files automatically by file extension.
- use BATS_TEST_DIRNAME variable to compute location of COMMAND.

Ref #2.


ahmetb:
cc: @feloy

ahmetb:
also:

I'm adding `echo "$output">&2` before a check that might fail the test, so that we can debug the test by its output. Otherwise all we'll get is an output like:


```
$ bats test/
1..4
not ok 1 --help should not fail
# (in test file test/kubectx.bats, line 7)
#   `[ "$status" -eq 0 ]' failed
not ok 2 -h should not fail
# (in test file test/kubectx.bats, line 12)
#   `[ "$status" -eq 0 ]' failed
not ok 3 --help should not fail
# (in test file test/kubens.bats, line 7)
#   `[ "$status" -eq 0 ]' failed
not ok 4 -h should not fail
# (in test file test/kubens.bats, line 12)
#   `[ "$status" -eq 0 ]' failed
```

PR #110, opened 2018-12-23
add kubectx and kubens as artifacts to tagged releases

maorfr:
This PR adds a `deploy` stage to `.travis.yml` which adds kubectx and kubens to a tagged release

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,12 @@ before_install:
   - sudo chmod +x /usr/bin/kubectl 
 script:
   - bats test/
+deploy:
+  provider: releases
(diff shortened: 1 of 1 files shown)


maorfr:
I signed it!

ahmetb:
Why do you need this while you can download the scripts directly from the tree? :)

maorfr:
Force of habbit, I guess. Thought it would be a nice touch...

You can close if you wish :)
Great tool, btw! Thanks.

ahmetb:
I'll close. I don't think we need this right now. If we move to language like Go in the future, we'll revisit.

PR #111, opened 2018-12-24
Add more kubectx tests

feloy:


test/common.bash
```diff
@@ -0,0 +1,32 @@
...
+setup() {
+  export KUBECONFIG=$(mktemp)
+  export XDG_CACHE_HOME="$(mktemp -d)"
+}
+
+# bats teardown function
+teardown() {
+  rm -f $KUBECONFIG
+  rm -f $XDG_CACHE_HOME/kubectx
+  rmdir $XDG_CACHE_HOME
```
ahmetb: you can combine all these into one `rm -rf $XDG_CACHE_HOME`, if you put kubeconfig in that dir, too.

test/common.bash
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bats
+
+# bats setup function
+setup() {
+  export KUBECONFIG=$(mktemp)
```
ahmetb: I don't understand why you don't do this like

    KUBECONFIG="$XDG_CACHE_HOME/kubeconfig"

you're gonna get a random dir every time anyway.

test/common.bash
```diff
@@ -0,0 +1,32 @@
...
+teardown() {
+  rm -f $KUBECONFIG
+  rm -f $XDG_CACHE_HOME/kubectx
+  rmdir $XDG_CACHE_HOME
+}
+
+# wrappers around "kubectl config" command
+
+add_cluster() {
+  kubectl config set-cluster ${1}
```
ahmetb: consider putting "" around all variable references, this prevents splitting by space.

test/kubectx.bats
```diff
@@ -2,12 +2,104 @@
...
+
+  run ${COMMAND} user2@cluster1
+  echo "$output"
+  [ "$status" -eq 0 ]
+  echo "$(get_context)"
+  [[ "$(get_context)" = "user2@cluster1" ]]
+}
+
+@test "create two contexts and switch between contexts" {
+  add_cluster cluster1
```
ahmetb: I feel like we are gonna repeat this 2-context kubeconfig file a lot. I wish we had a way of automatically getting that.

test/kubectx.bats
```diff
@@ -2,12 +2,104 @@
...
+  add_user user1
+  add_user user2
+  add_context user1@cluster1 user1 cluster1
```
ahmetb: I don't think you should modify $KUBECONFIG file in tests.

You can consider doing a `cp testdata/kubeconfig "$(mktemp)"` for each test, and use that copy. Otherwise we'd be mutating kubeconfig, which can cause problems when we reorder tests. Ideally each test should be idempotent/isolated.

ahmetb:
Looking good. Thanks! 🙌

PR #113, opened 2018-12-25
Complete tests for kubectx

feloy:


ahmetb:
cc: @doodlesbykumbi your review is appreciated!

test/kubectx.bats
```diff
@@ -25,7 +35,14 @@ load common
   [[ "$output" = "error: No previous context found." ]]
 }
 
-@test "create one context and list contexts" {
+@test "list contexts when no kubeconfig exists" {
```
ahmetb: this is currently not immune to my compute having a ~/.kube/config

please deliberately do `KUBECONFIG=./non-existing-file kubectx` here. it should work the same way.

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+  use_config config2
+
+  run ${COMMAND} "new-context=user1@cluster1"
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"new-context"* ]]
```
ahmetb: can you also test `user1@cluster1` doesn't exist anymore

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND} new-context=.
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"user1@cluster1"* ]]
```
ahmetb: can you also test the current context name doesn't exist anymore?

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+  [[ "$output" = "user1@cluster1" ]]
+}
+
+@test "delete non existent context" {
+  use_config config2
+
+  run ${COMMAND} -d "unknown-context"
+  echo "$output"
+  [ "$status" -eq 1 ]
+}
```
ahmetb: can you please add tests for deleting multiple contexts, too?

I'm actually not sure what its behavior when you `-d a b c` but `b` doesn't exist, so does `c` still get deleted... It'd be a good behavior to test.

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = "user1@cluster1" ]]
+}
+
+@test "delete non existent context" {
+  use_config config2
```
ahmetb: probably can use config1 here (try using simple ones when you can)

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND} -d .
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = "user1@cluster1" ]]
```
ahmetb: instead of testing "what's left in the output" (fragile, because we can change what's in `config2`), test "what shouldn't be in the output".

so please test the deleted context no longer shows up in the output. syntax is: `[[ ! "$output = *"name"* ]]`

test/kubectx.bats
```diff
@@ -87,3 +104,81 @@ load common
...
+
+  run ${COMMAND} "new-context=user1@cluster1"
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"new-context"* ]]
+  [[ "$output" = *"user2@cluster1"* ]]
```
ahmetb: isn't the regexp syntax supposed to be `=~`?
why does this even work 🤔

ahmetb: hmm interesting I guess this is just bash glob feature without regexps. in that case it's fine.

```
[[ "ahmet" = *"ahm"* ]]; echo $?
0
```

test/kubectx.bats
feloy: there is a export KUBECONFIG=... in setup(), in common.bash, I understand that it is equivalent

ahmetb: Ah right, but doesn’t that get changed or something if we reorder the tests?

It would be good to explicitly set a kubeconfig to a nonexisting file to test this without side effects of the outside world.

ahmetb:
Please ping me when it's ready to review again. I'm super happy of this work!

test/kubectx.bats
```diff
@@ -87,3 +104,110 @@ load common
...
+  use_config config2
+
+  run ${COMMAND} -d "user1@cluster1"
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = "user2@cluster1" ]]
```
ahmetb: similarly here, please first test the deleted context doesn't exist
I don't think testing "what's left" isn't as crucial.

feloy:
I rebased my commits so they are attached to my github account now.

test/kubectx.bats
feloy: I would prefer to also test what's left, to be sure all is not deleted

ahmetb: yes that makes sense.

ahmetb:
Thanks a lot! This brings us to 19 tests for kubectx alone.

doodlesbykumbi:
Hey y'all. Sorry, quite late to the game was thoroughly AFK over the new year break. This is amazing work. Well done @feloy!

ahmetb:
Hmm looks like we have 24 tests alone in this file. but the test output showed 19 for some reason (both kubectx and kubens) https://travis-ci.org/ahmetb/kubectx/builds/474502614?utm_source=github_status&utm_medium=notification

```
$ bats test/
1..19
ok 1 no kubectl detected
ok 2 --help should not fail
ok 3 -h should not fail
ok 4 switch to previous context when no one exists
ok 5 list contexts when no kubeconfig exists
ok 6 get one context and list contexts
ok 7 get two contexts and list contexts
ok 8 get two contexts and select contexts
ok 9 get two contexts and switch between contexts
ok 10 get one context and switch to non existent context
ok 11 rename context
ok 12 rename current context
ok 13 delete context
ok 14 delete current context
ok 15 delete non existent context
ok 16 delete several contexts
ok 17 delete several contexts including a non existent one
ok 18 --help should not fail
ok 19 -h should not fail
```

doodlesbykumbi:
perhaps we can split the invocations ?
```
script:
  - echo "kubectx tests:"; bats test/kubectx.bats
  - echo "kubens tests:"; bats test/kubens.bats
```

doodlesbykumbi:
@ahmetb i'm confused there's only 19 instances of `@test` in the repo. where's 24 coming from ?

ahmetb:
> where's 24 coming from ?

my bad. I was ctrl+f ing the diff view.

PR #116, opened 2019-01-02
split bats test invocation by binary

doodlesbykumbi:

PR #117, opened 2019-01-03
add more cli tests for kubens

doodlesbykumbi:


ahmetb:
I think you pushed to the branch of #116 so I can't merge that now.

kubens
```diff
@@ -118,6 +118,8 @@ choose_namespace_interactive() {
 set_namespace() {
   local ctx prev
   ctx="$(current_context)" || exit_err "error getting current context"
+  [[ -z "$ctx" ]] && exit_err "current context is empty"
```
ahmetb: I can't imagine a case where this returns empty string.  Can you clarify? Ideally the previous line should fail.

kubens
```diff
@@ -173,14 +175,19 @@ swap_namespace() {
 }
 
 main() {
-  if hash kubectl 2>/dev/null; then
+  set +u
```
ahmetb: don't do this. the many `set -u`s you have below clearly shows it'll cause trouble someday when we forget to add `set -u` to a new code path. :)

instead use a one-off default like:

    "${KUBECTL:-}"

in the IF statement.

doodlesbykumbi:
@ahmetb i closed #116 and deleted the branch then created a brand new branch with all the changes. you should be good to merge from here.

kubens
```diff
@@ -173,14 +175,19 @@ swap_namespace() {
...
+  if [[ ! -z "$KUBECTL" ]]; then
```
ahmetb: I believe this is equivalent to `-n`. :)

mock-kubectl
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
ahmetb: move this to `test/`

mock-kubectl
```diff
@@ -0,0 +1,23 @@
...
+
+[[ -n $DEBUG ]] && set -x
+
+set -eou pipefail
+
+if hash kubectl 2>/dev/null; then
```
ahmetb: I don't understand why this code block is here. 

I think we should override KUBECTL only for calls that fail without a cluster (like `kubectl get namespace`); not the entire test suite.

that would simplify both the mock-kubectl script, and makes it easier to reason about the tests.

mock-kubectl
```diff
@@ -0,0 +1,23 @@
...
+  KUBECTL=kubectl
+elif hash kubectl.exe  2>/dev/null; then
+  KUBECTL=kubectl.exe
+else
+  set -u
+  echo >&2 "kubectl is not installed"
+  exit 1
+fi
+
+if [[ $@ == *' namespaces '* ]]; then
```
ahmetb: we can probably be more specific like `$@ == "get namespaces"`

test/common.bash
```diff
@@ -17,6 +17,29 @@ use_config() {
...
+}
+
 get_context() {
-    kubectl config current-context
+  kubectl config view -o=jsonpath='{.current-context}'
+}
+
+exit_err() {
+  echo >&2 "${1}"
+  exit 1
```
ahmetb: I'm not sure if `exit 1` is the way to fail from a bats test.

Can you just try making the last command `false` here, and see if it fails properly? I suspect if all `set -e`s are correct, this failure should propagate.

test/kubens.bats
```diff
@@ -13,3 +16,90 @@ COMMAND="${BATS_TEST_DIRNAME}/../kubens"
...
+}
+
+@test "switch to previous namespace when none exists" {
+  use_config config1
+  switch_context user1@cluster1
+
+  run ${COMMAND} -
+  echo "$output"
+  [ "$status" -eq 1 ]
+  [[ "$output" = "error: No previous namespace found for current context." ]]
```
ahmetb: I wouldn't compare error messages, these are prone to change a bit frequently.

doodlesbykumbi: that may be true but i think it's still worth asserting on the output as much as on the exit status code. 
the test case seems empty to me without asserting on the output

test/common.bash
doodlesbykumbi: noted but turns out this wasn't necessary at all

mock-kubectl
doodlesbykumbi: updated

mock-kubectl
doodlesbykumbi: removed all this junk. it was unnecessary after all.

mock-kubectl
doodlesbykumbi: noted and done

kubens
doodlesbykumbi: 🤦‍♂️... thanks

doodlesbykumbi:
@ahmetb addressed your comments and replied where it made sense to do so

kubens
doodlesbykumbi: this was also not necessary

test/kubens.bats
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bats
 
 COMMAND="${BATS_TEST_DIRNAME}/../kubens"
+export KUBECTL="$BATS_TEST_DIRNAME/../test/mock-kubectl"
```
ahmetb: If we override KUBECTL _only_ when we need the "get namespaces" functionality, would that pretty much be _all_ test cases?

test/kubens.bats
```diff
@@ -13,3 +16,90 @@ COMMAND="${BATS_TEST_DIRNAME}/../kubens"
   echo "$output">&2
   [ "$status" -eq 0 ]
 }
+
+@test "list namespaces when no kubeconfig exists" {
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq "1" ]
```
ahmetb: I think we should use `[[` for consistency, no reason not to. 

also you used `"1"` here vs `0` above, please don't quote the numeric value.

test/mock-kubectl
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+[[ -n $DEBUG ]] && set -x
+
+set -eou pipefail
+
+if [[ $@ == *'get namespaces'* ]]; then
+  echo "default"
```
ahmetb: let's call these ns1 and ns2. no need to stick to original names (hard to follow in test cases above), and no need to have 3.

kubens
```diff
@@ -173,7 +173,9 @@ swap_namespace() {
 }
 
 main() {
-  if hash kubectl 2>/dev/null; then
+  if [[ "${KUBECTL:-}" ]]; then
```
ahmetb: I don't know what `[[ "$var" ]]` test does (even though manpage says it checks string isn't null), I recommend you stick to -n or -z here.

Not sure what the next line saying `true` does either.

I recommend if you rewrite this like

```
if [[ -z "${KUBECTL:-}" ]] then;
  if hash ..
  elif ..
  else ..
  fi

fi
```

test/kubens.bats
```diff
@@ -13,3 +16,90 @@ COMMAND="${BATS_TEST_DIRNAME}/../kubens"
...
+
+  run ${COMMAND}
+  echo "$output"
+  [ "$status" -eq 0 ]
+  [[ "$output" = *"default"* ]]
+  [[ "$output" = *"kube-public"* ]]
+  [[ "$output" = *"kube-system"* ]]
+}
+
+@test "switch to existent namespace" {
```
ahmetb: existing*

test/kubens.bats
doodlesbykumbi: it seems that way for now. `get namespaces` is the only call to `kubectl` that requires a cluster

test/docker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ruby:2.3.8
```
ahmetb: I think you checked these in accidentally :)

ahmetb: Ah, no thanks, please don't include changes that weren't discussed. :) Unless you're on windows, it's pretty easy to install bats. So I don't think we should add scripts and dockerfiles almost as long as the tests to run the tests.

doodlesbykumbi: Fair enough. Pruned :)

ahmetb:
Thanks these are looking good now! 🙏

Issue #115, opened 2018-12-28
Proposal: write kubectx/kubens in Go


ahmetb:
Most people don't realize `kubectx` and `kubens` are small bash scripts today.

Bash works fine and has brought us this far. Since we mostly shell out to kubectl for config processing and the project is minimalist enough, this has been working well.

## Problems with using bash

However, for long-term maintenance, using bash scripting will be a problem:

- **Bash is not cross-platform, hence divides the community:** Many users quote they can't use kubectx on windows. So some of them have gone on and wrote their "clones" of kubectx that work on Windows. 

  Some examples of fragmentation from the ecosystem:

  - Kubectxwin: https://github.com/thomasliddledba/kubectxwin
  - kubectx-rs (Rust): https://github.com/FGRibreau/kubectx-rs
  - kubectl-ctx (Go, a kubectl plugin): https://github.com/postfinance/kubectl-ctx
  - kubectx (poorly named, Node.js) https://github.com/FGRibreau/kubectx, another one in nodejs: https://github.com/WoLfulus/kubee
  - someone emailed me yesterday asking if they can use "kubectx" name and publish a Go implementation
  - Issue to support windows: #29
  - kx (Go, related): https://github.com/kylie-a/kx

  This is pointing to clear community interest for a kubectx that works on windows, despite bash-on-Windows and Windows Subsystem for Linux (WSL) are now widely available.

- **People don't want contribute to bash scripts:** I like bash alright, most people don't want to code in bash. This is probably dramatically reducing the contributor and maintainer prospects pool drastically.

- **Hard to reason about code:** Some parts of the kubectx code are about to get out of control, such as the flags parsing/handling.

## Problems with switching to Go

- **Rewrite:** We need to rewrite. Arguably the code would be longer, but more maintainable. We already have some integration/CLI tests effort happening in #2, which should help test post-rewrite.

- **Larger binary size:** We're going from 5kb to ~5-10 mb in exchange for platform portability.

- **More difficult to support POSIX platforms:** If you had bash, kubectx would work fine today. But if we rewrite in Go, supporting architectures like ppc or arm (<1% users) , would require making separate Go binary releases. Bash doesn't have this problem.

-----

I'm 51% inclined to rewrite in Go (and release v1.0). I appreciate kubectx user community can chime in with their thoughts on this. ❤️

arush-sal:
Let me know if any help is needed.

bitfield:
[...]

PR #119, opened 2019-01-09
Add ZSH framework support

unixorn:
This file enables loading the repository as a plugin by oh-my-zsh compatible frameworks.

The big advantage to this is that frameworks like [antigen](https://github.com/zsh-users/antigen) and [zgen](https://github.com/tarjoilija/zgen) will handle automatically cloning the repository, adding the checkout's directory to the user's `$PATH`, and also will periodically `git pull` to keep them up to date without the user having to remember to do it.

### DIFF
--- a/kubectx.plugin.zsh
+++ b/kubectx.plugin.zsh
@@ -1,0 +1,7 @@
+#!/usr/bin/env zsh
+#
+# Enables this repo to be autoloaded by ZSH frameworks that support
+# oh-my-zsh format plugins.
+
+KUBECTX_DIR="$(dirname $0)"
+export PATH=${PATH}:${KUBECTX_DIR}


ahmetb:
Oh this is a good idea.

I wonder if we should move this to contrib/ and use dirname(dirname($0)) instead.

ahmetb:
But also I don't think oh-my-zsh upgrades plugins (not desirable) and we may move away from kubectx/kubens being bash-scripts (#115) which wouldn't work with this.

unixorn:
oh-my-zsh doesn't, but most of the other ZSH frameworks do. It doesn't matter what language `kubectx` and `kubens` are, the plugin code I added just adds the repo root directory to the user.

Thanks for merging.

ahmetb:
I am not sure this will work when we move to Go and no longer provide an executable in the source code.

If I merged this PR, you could do:

1. clone this repo to /foo/bar
2. add /foo/bar/kubectx.plugin.zsh to zsh plugins list
3. restart and /foo/bar will be in your PATH

When we switch to Go, Step 1 will no longer get you kubectx. I hope I'm getting this right. Does this make sense?

I'm trying to understand which platform does this help distribute. Are you focusing  oh-my-zsh users on macOS or Linux? How would the zsh-plugin based installation instructions look like?

unixorn:
When it switches to go, this will have to be updated.

I'd probably add a check to see if `kubectx` is in the path, and if not, do a `go get xyzzy` to build it.

My goal was to make it easier to install and keep current on both linux and macOS since I use macOS on my laptop and linux on my servers. I currently use the same plugin list with `zgen` on both, and let `zgen` handle cloning them and keeping them up to date.

ahmetb:
We have a well-established packaging for macOS through Homebrew.

But Linux has been a trouble. So far we're just telling people to copy the scripts into somewhere in their PATH.

If we switch to go as described in #115, this opens possibilities of using package managers like apt/yum (deb/rpm) in addition to Homebrew through projects like https://goreleaser.com/.

I think adding something to PATH shouldn't require that to be packed as a ZSH plugin. This is not how most things are installed. So I'm tempted to close this.

unixorn:
The current situation on linux is that users have to manually do a git checkout, manually add the scripts to their path, and manually remember to check back for updates, which doesn't let us notify them when a bug fix is released or the go rewrite is complete. And I have to do that on every single server I want `kubectx` on.

How about merging it until https://github.com/ahmetb/kubectx/issues/115 is implemented? It is a lot easier to have the framework handle checking the repo out, adding the directory to $PATH, and most importantly, keeping it up to date. Then I can log into one of my servers, `zgen` sees my plugin list is updated, and it handles everything for me.

I guess my point is that this works on linux _now_, and we don't know how long it will take to do the go rewrite. It isn't going to hurt anything to add plugin support, especially since it makes it easier for us to notify existing users of changes since most of the existing ZSH frameworks will automatically keep their checkout of the code up to date. Oh-my-zsh users will have to update manually, but they already have to update all their external plugins manually.

When it gets rewritten in go, we can update the ZSH plugin code to display a upgrade message like "`kubectx` has been rewritten in go. Please read https://github.com/ahmetb/kubectx/how_to_update_to_go_version.md for details on how to upgrade your installation."

ahmetb:
> most of the existing ZSH frameworks will automatically keep their checkout of the code up to date. Oh-my-zsh users will have to update manually, but they already have to update all their external plugins manually.

What are some other ZSH frameworks? oh-my-zsh is the only one I know (and probably the most popular), and as you said, it doesn't know anything about plugin updates.

unixorn:
I've got a full list of frameworks and plugins at [awesome-zsh-plugins](https://github.com/unixorn/awesome-zsh-plugins), but I've personally just used [antigen](https://github.com/zsh-users/antigen) and [zgen](https://github.com/tarjoilija/zgen).

I prefer zgen because it renders an init.zsh to source only when you change your plugin list, so it loads much faster. Antigen will also handle cloning and updating plugin repos like zgen does, but because it examines your plugin list every time you start a terminal, it's a little slower. They both support using oh-my-zsh plugins directly from checkouts of the omz repository.

ahmetb:
This is a bit too exotic and I'm inclined to not accept this contribution since I can't maintain it and not many software are distributed like this.

PR #123, opened 2019-01-30
Fixed fzf issue on ubuntu with kubectl query

shaheinm:
I use this tool on macOS (installed via homebrew) with no issues, but when I cloned the repo on my ubuntu machine, the `$SELF_CMD` feed to `fzf` didn't work (I still haven't figured out _exactly_ why...).  The more I looked at it, the more I wondered why not just feed the results of the `kubectl` query directly to `fzf` and select from there, and when I changed it, it worked with no issues on ubuntu or macOS.

I also made a change to the `fzf` output, opting out of fullscreen mode and having it drop down from the current prompt.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -21,7 +21,6 @@
 set -eou pipefail
 IFS=$'\n\t'
 
-SELF_CMD="$0"
 KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
 
 usage() {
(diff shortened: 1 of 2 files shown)


ahmetb:
Instead of sending a nontrivial fix, can you please open an issue first with the behavior you’re seeing? You can set DEBUG=1 to get more info.

shaheinm:
Apologies, fair enough.

It appears as though my issue is tied to an Ubuntu kernel update released recently (https://github.com/junegunn/fzf/issues/1486).  I'll wait to see what happens there before submitting the issue (since it doesn't appear to actually be an issue in `kubectx`).  Thanks!

ahmetb:
Thanks!

PR #124, opened 2019-01-30
Add version flag for kubectx and kubens

cbeneke:
When having multiple versions of kubectx installed it is currently not possible to differentiate between the versions except with looking into the code and comparing the recent changes (or checking the git tags). In my head it is convenient to have a version flag on each script. This PR adds a (_very_ basic) `--version` flag.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -23,6 +23,7 @@ IFS=$'\n\t'
 
 SELF_CMD="$0"
 KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
+KUBECTX_VERSION=""
 
 usage() {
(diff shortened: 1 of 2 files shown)


ahmetb:
Instead of implementing it, you could just make a proposal issue where we discuss on how to implement this. :) Right now I’m -1 for hardcoding versions which then someone has to go update on each release, which they may forget to do so.

cbeneke:
Sorry, sure thing. I added an issue regarding this PR. Didn't want to put too much logic into it. The 'might be forgotten' parts is true though^^

cbeneke:
Closing here in favor of the issue.

PR #128, opened 2019-02-04
failed to read tty error with fzf on ubuntu 

emreodabas:
kubectx not working with fzf on ubuntu cause of ubuntu kernel bug -> https://github.com/junegunn/fzf/issues/1486

Using fzf with -1 parameter is a workaround for this bug and also give auto select if list has only one value.


PatelParas:
Remove `/home/user/.fzf/bin` from $PATH and your good. temporary fix.

PR #130, opened 2019-02-07
Proposal: Automatically select the namespace if there is only one option after fuzzy search

dineshba:
This PR will make `kubectx` and `kubens` select namespace **directly** *if there is a single result* based on fuzzy match.

**Note:** This will work only when `fzf` is enabled

**Example:**
Lets say we have three namespaces.

    $ KUBECTX_IGNORE_FZF=false kubens
    default
    kube-public
    kube-system

With this PR,

    kubens def # will change to default namespace
    kubens sys # will change to kube-sytem namespace
    kubens pub # will change to kube-public namespace
    kubens ksy # will change to kube-system namespace
    kubens kube # will take you to interactive session where we have to select between kube-public and kube-system
Similarly, it will work for `kubectx`

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -106,7 +106,7 @@ choose_context_interactive() {
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --query "$1" --select-1 || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1
@@ -184,7 +184,7 @@ main() {
 
   if [[ "$#" -eq 0 ]]; then
     if [[ -t 1 &&  -z "${KUBECTX_IGNORE_FZF:-}" && "$(type fzf &>/dev/null; echo $?)" -eq 0 ]]; then
-      choose_context_interactive
+      choose_context_interactive "$*"
     else
       list_contexts
     fi
@@ -211,7 +211,11 @@ main() {
     elif [[ "${1}" =~ (.+)=(.+) ]]; then
       rename_context "${BASH_REMATCH[2]}" "${BASH_REMATCH[1]}"
     else
-      set_context "${1}"
+      if [[ -t 1 &&  -z "${KUBECTX_IGNORE_FZF:-}" && "$(type fzf &>/dev/null; echo $?)" -eq 0 ]]; then
+        choose_context_interactive "${1}"
+      else
+        set_context "$1"
+      fi
     fi
   else
     usage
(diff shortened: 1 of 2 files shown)


dineshba:
I signed it!

ahmetb:
⚠️  **NOTE:** Please make a proposal next time you write the code for it so your work is not discarded when we don't agree on a change.

----

As for your change, I'm against these sort of fuzzy behavior. We incorporated `fzf` for two reasons:

- it gives you ability to scroll between options using arrow keys
- you can filter down options by writing some queries

I am not willing to add more features beyond simple integration with fzf. Here's why:
- This feature would **only** work with fzf. And it would complicate documentation.
- It creates different behavior for the same command when fzf is installed vs not.  This is not desirable.
- FZF employs a fuzzy search, it's not a substring search.

If we don't make this change, you'll basically be typing only more keystroke (you can type those queries while in fzf interactive mode, and hit Return).

For these reasons I'm currently -2 for this change unless you have counter-arguments to change my mind.

dineshba:
Sure @ahmetb. I am relatively new to this `OSS community`. From next time, I will create an issue describing the request/proposal and if everyone agreed, we can start to implement the idea.

Few points that I feel this feature is useful:
- I agree that it will give a **different behavior for the same command** when fzf is installed vs not. I am seeing this very similar to command `kubens`. Without fzf, `kubens` will list the namespaces and with fzf, `kubens` will take us to the interactive mode where we can filter, navigate and select  a namespace (interactive mode makes our life easier to select a namespace)
- With this we can easily change the namespace with one command + less keystrokes 
   `kubens kube-systems` vs `kubens sys`.
- If there are  multiple namespaces, we have to enter into interactive and query it and select it. Here if we remember the name vaguely(fuzzy), we can select the namespace in one command (no interactive, no extra typing, no extra enter)

So I feel this will increase the **productivity** of the user. And I totally agree we have to add more docs to explain this feature(I would love to do that)

ahmetb:
> * With this we can easily change the namespace with one command + less keystrokes
>   `kubens kube-systems` vs `kubens sys`.

It's still 1 command, and as I said 1 keystroke fewer is not a good argument.



> * If there are  multiple namespaces, we have to enter into interactive and query it and select it. Here if we remember the name vaguely(fuzzy), we can select the namespace in one command (no interactive, no extra typing, no extra enter)

Here's a better counter-argument: A lot of people run `kubectx` or `kubens` to see the namespaces available. And most users start with 1 context. Therefore, making a choice instead of switching breaks that behavior.

If I'm running `kubectx` it means:
- (in interactive mode) I want to see a list of {contexts,namespaces} and (maybe) choose one.
- (non-interactive mode) I want to see a list of {contexts,namespaces}

Many people run kubectx+fzf and hit <kbd>Ctrl</kbd>+<kbd>C</kbd> to exit, because they came there to see the list. I don't think we should make this command have side-effects when there's only 1 item to show.

PR #138, opened 2019-03-29
improved fzf usage

caarlos0:
this passes down `$1` to `fzf`, so it may pre-choose something for you.

It looks like this:

Kapture 2019-03-29 at 13 59 41

IMHO it is more practical because I can type `kx something` and even if its not the complete name it may go to what I need already. 🥇

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -106,7 +106,7 @@ choose_context_interactive() {
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi -q "$1" -1 || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1
@@ -184,7 +184,7 @@ main() {
 
   if [[ "$#" -eq 0 ]]; then
     if [[ -t 1 &&  -z "${KUBECTX_IGNORE_FZF:-}" && "$(type fzf &>/dev/null; echo $?)" -eq 0 ]]; then
-      choose_context_interactive
+      choose_context_interactive ""
     else
       list_contexts
     fi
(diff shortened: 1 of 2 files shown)


caarlos0:
(if you like the idea I'll work on the tests)

ahmetb:
I really like it, I gotta try this myself. How are you planning to test it?

ahmetb:
On a second look, I'm a bit skeptical since `kubectx <NAME>` could be used in the scripts, so it probably shouldn't trigger "interactive" selection. I think this breaks our contract. :(

caarlos0:
> On a second look, I'm a bit skeptical since `kubectx <NAME>` could be used in the scripts, so it probably shouldn't trigger "interactive" selection. I think this breaks our contract. :(

🤔

I wouldn't really use `kubectx` in a script, would rather use `kubectl` for that.. but I get your point... maybe disable it by default with an environment variable or something?



> I really like it, I gotta try this myself. How are you planning to test it?

don't know yet, just seen tests are failing but haven't looked into them yet 🤔 just wanted to propose the change and see if you like it :)

ahmetb:
> I wouldn't really use `kubectx` in a script, would rather use `kubectl` for that

exactly my thoughts. I repeated this 193517093 times in this very issue tracker, but people are still using kubectx in scripts, that's why I kind of gave up on that fight. 🤷‍♂️ 

This PR changes a very basic behavior fundamentally _when_ fzf is installed (most users don't have fzf), so it might be OK. Technically it's equal number of keystrokes with or without your PR.

Let's think about it for a while.

caarlos0:
> exactly my thoughts. I repeated this 193517093 times in this very issue tracker, but people are still using kubectx in scripts, that's why I kind of gave up on that fight. 🤷‍♂️

I understand haha



> This PR changes a very basic behavior fundamentally _when_ fzf is installed (most users don't have fzf), so it might be OK. Technically it's equal number of keystrokes with or without your PR.

yeah, it saves time only if you have things like `some-really-big-context` and then you do `kubectx big` and it goes directly without even showing `fzf`'s interface... 



> Let's think about it for a while.

yes, not problem!

Thanks for the great tool BTW 🚀

ahmetb:
> yeah, it saves time only if you have things like `some-really-big-context` and then you do `kubectx big` and it goes directly without even showing `fzf`'s interface...

this idea came up before and I was not in favor: https://github.com/ahmetb/kubectx/pull/130

ahmetb:
I'll close this as there might be others relying on existing behavior for "kubectx $1" to be an exact match. We have shell completion that provides exact values for the argument.

PR #139, opened 2019-04-03
kubectx rename check if old_name is a valid ctx

### LINKED ISSUE #136: FeatureRequest: Prevent accidental deletion of context when swapping the new and old context in kubectx
To **prevent accidental deletion of a context** when swapping the old and new context parameters on a rename operation check at first if the `oldContext` exists.

Use Case:
```
# correct usage
kubectx newContext=oldContext
# wrong usage which deletes oldContext
kubectx oldContext=newContext
```

Currently in case of wrong usage:
```
user:~$ kubectx oldContext=newContext
Context "oldContext" exists, deleting...
error: cannot rename the context "nexContext", it's not in /Users/user/.kube/config
```

Preferred in case of wrong usage: (no context gets deleted)
```
user:~$ kubectx oldContext=newContext
error: The context "nexContext", is not in /Users/user/.kube/config
```

ahmetb:
Without this safeguard, when user runs `kubectx NEW_NAME=OLD_NAME` where
NEW_NAME is an existing context but OLD_NAME isn't, we end up deleting NEW_NAME
and not doing any renames (because OLD_NAME is not found).

Fixes #136.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -148,6 +148,11 @@ rename_context() {
     old_name="$(current_context)"
   fi
 
+  if ! context_exists "${old_name}"; then
+    echo "error: Context \"${old_name}\" not found, can't rename it." >&2
+    exit 1
+  fi
+
   if context_exists "${new_name}"; then
     echo "Context \"${new_name}\" exists, deleting..." >&2
     $KUBECTL config delete-context "${new_name}" 1>/dev/null 2>&1


RonaldFindling:
Looks fine for me --> local tryout throws the expected error and doesn't delete my old context :)

RonaldFindling:
Looks fine for me --> local tryout throws the expected error and doesn't delete my old context :)

PR #141, opened 2019-04-04
add a switch --current for printing current namespace

nicholasklem:
Not sure this belongs in this utility, but I need "current namespace" for other scripts I am using, so here is a patch!

Feel free to delete if not appropriate, or rename it if wanted.


nicholasklem:
Whitespace errors!?!!

PR #143, opened 2019-04-05
Adaptative height for choice

eplanet:
This is a modification I made locally to avoid having a fullscreen `fzf` and instead only an adaptative height to the number of available contexts. The 2 other lines are the current choice and the counter.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -21,8 +21,11 @@
 set -eou pipefail
 IFS=$'\n\t'
 
-SELF_CMD="$0"
-KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
+KUBECTX_PREVIOUS_VALUE="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
+KUBECTX_CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/cache/kubectx"
+[[ ! -d ${KUBECTX_CACHE_DIR} ]] && mkdir -p ${KUBECTX_CACHE_DIR}
+CONTEXT_LIST_CACHE="${KUBECTX_CACHE_DIR}/context_list"
+SELF_CMD="cat ${CONTEXT_LIST_CACHE}"
 
 usage() {
   cat <<"EOF"
@@ -50,7 +53,7 @@ current_context() {
 }
 
 get_contexts() {
-  $KUBECTL config get-contexts -o=name | sort -n
+  $KUBECTL config get-contexts -o=name | sort -n | tee ${CONTEXT_LIST_CACHE}
 }
 
 list_contexts() {
@@ -84,8 +87,8 @@ list_contexts() {
 }
 
 read_context() {
-  if [[ -f "${KUBECTX}" ]]; then
-    cat "${KUBECTX}"
+  if [[ -f "${KUBECTX_PREVIOUS_VALUE}" ]]; then
+    cat "${KUBECTX_PREVIOUS_VALUE}"
   fi
 }
 
@@ -94,7 +97,7 @@ save_context() {
   saved="$(read_context)"
 
   if [[ "${saved}" != "${1}" ]]; then
-    printf %s "${1}" > "${KUBECTX}"
+    printf %s "${1}" > "${KUBECTX_PREVIOUS_VALUE}"
   fi
 }
 
@@ -102,11 +105,23 @@ switch_context() {
   $KUBECTL config use-context "${1}"
 }
 
+context_list_count() {
+  if [[ -f ${CONTEXT_LIST_CACHE} ]]
+  then
+    echo $(($(cat ${CONTEXT_LIST_CACHE} | wc -l) + 2))
+  else
+    echo "100%"
+  fi
+}
+
 choose_context_interactive() {
+  list_contexts > /dev/null
+  HEIGHT="$(context_list_count)"
+
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --height ${HEIGHT} || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1


ahmetb:
Oh nice. This is gonna make it a little slower due to extra call to kubectx (and therefore to kubectl).  Also some users have >100 contexts, how is it gonna be handled here?

eplanet:
Yeah I wonder if we could cache the list of contexts in `~/.kube/contexts` and refresh it only once in that case?

For more than 100 contexts, it just doesn't change the existing behavior of going full screen (`fzf` rocks...).

ahmetb:
#### Latency

Unfortunately caching might not work well here (as new contexts might be added/removed outside this tool, or KUBECONFIG env var might be set). This also complicates the code heavily.

I just tested and I think latency-wise it's not bad. But I think we can improve. See this:

- `kubectl config get-contexts -o=name | wc -l` takes ~90-100 ms
- `kubectx | wc -l` takes 190-240ms.

There's a 2x difference there (I'm not sure why, but it's kinda normal). So I think we should directly shell out to kubectl to count the contexts.

#### Evaluation of cmd

If we are doing stuff like ` $(( $(${SELF_CMD} | wc -l) + 2))` we should probably have `eval` around SELF_CMD as it could have multiple args (but still a single string value), like `SELF_CMD="seq 0 200"`, which would only work if you `eval` it.

So `$(( $(eval ${SELF_CMD} | wc -l) + 2))` is better.

#### kubens

We should probably make the same change in `kubens` at some point for symmetry. (Not urgent.) However, the latency problem is gonna be a big problem  for kubens.

Since `kubens` actually makes an API request, the first request can be as slow as 5 seconds or more (at least in GKE case) when token is expired. Even when there's a fresh token, my experimentation shows ~400-550ms for `kubectl get namespaces` request for a GKE cluster.

It almost feels like maybe fzf should have a feature that determines the size (+ fullscreen or not) based on the $FZF_DEFAULT_COMMAND's result. Maybe that's a feature request to fzf, though I'm not sure it's doable.

eplanet:
I just added a commit with some basic caching. Since it would duplicate the output of `${SELF_CMD}`, don't you think it would work well in all cases? I'm not very familiar with `KUBECONFIG`.

I think your comment on `eval` doesn't apply if we use caching. And I can do a similar thing for `kubens` if you like the approach.

ahmetb:
Yeah let me think on caching a little bit. There are problems like if we're running two simultaneous instances of kubectx with different KUBECONFIG env vars, they will cache different things. But I'm not sure people run two `kubectx`es in 0.2s succession.

eplanet:
Oh I see what you mean. I only use one configuration so I didn't think about that. Worst case I suppose is that you have a funny height once, but I understand it might be misleading.

And by the way I realize that caching namespaces might be tricky as you can have a big difference in namespaces cardinality between contexts. But we could cache them per context, as you do with previous namespace.

kubectx
```diff
@@ -23,6 +23,7 @@ IFS=$'\n\t'
 
 SELF_CMD="$0"
 KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
+KUBECTX_CONTEXTS="${KUBECTX}_contexts"
```
ahmetb: I think we should do:

```sh
CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/cache"
KUBECTX_CACHE_DIR="${CACHE_DIR}/kubectx"  # we should use this dir for $KUBECTX file too
context_list_cache="${KUBECTX_CACHE_DIR}/context_list"
```

kubectx
```diff
@@ -22,7 +22,10 @@ set -eou pipefail
 IFS=$'\n\t'
 
 SELF_CMD="$0"
-KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"
```
eplanet: I wonder if you want to move that file to the new path, if it exists, for backward compatibility?

kubectx
```diff
@@ -22,7 +22,10 @@ set -eou pipefail
...
+KUBECTX_CACHE_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/cache/kubectx"
+[[ ! -d ${KUBECTX_CACHE_DIR} ]] && mkdir -p ${KUBECTX_CACHE_DIR}
+KUBECTX_CACHE="${KUBECTX_CACHE_DIR}/kubectx"
```
eplanet: I'd actually rename that to `PREVIOUS_CONTEXT_CACHE`, since it's really used as a storage for previous context.

eplanet:
Could you have a look at the changes to validate it goes it the correct direction?

ahmetb:
Sorry got busy lately, I'll come back to this ASAP

kubectx
ahmetb: Yeah I think we should keep $KUBECTX the same, but use the new $KUBECTX_CACHE_DIR while determining $CONTEXT_LIST_CACHE path.

Besides, `~/.kube/cache` can be nuked occassionally, so it would be good to keep the last-context file outside `cache` dir.

eplanet: I restored it in its original path but renaming it `KUBECTX_PREVIOUS_VALUE`. Do you think we should stick to `KUBECTX` only?

kubectx
```diff
@@ -103,10 +106,12 @@ switch_context() {
 }
 
 choose_context_interactive() {
+  [[ -f ${CONTEXT_LIST_CACHE} ]] && HEIGHT=$(($(eval cat ${CONTEXT_LIST_CACHE} | wc -l) + 2)) || HEIGHT="100%"
```
ahmetb: 1. you don't need `eval` in `$(eval cat ...)`

2. please unfold `[[ -f ]] && ` as `if [..`

3. I think you should create a method named `context_list_count` and call it from here instead of embedding the file logic here.

4. I think there's a bug here. first time I call it (with no `.cache` dir), I get 100%, but the next time it works. It appears like the HEIGHT doesn't reflect the current count, but the previous run. This is a problem when I just added a new 5th context, but next time I run kubectx, height will be =4 (and so it won't show one of the contexts).

ahmetb: the solution to #4 might be to actually modify SELF_CMD to use command `cat $cache_file`.

because if we just computed the cache, it's probably cool.

eplanet:
Thanks for your feedback! I made required updates, tested successfully with height being correct both before and after the following changes:
- No cache existing
- Removing the cache
- Adding a context
- Removing a context

If you're OK with it, shall I rebase the PR or do you just squash it?

Please do not hesitate to share any further concern.

ahmetb:
Thanks a lot. Don’t worry about squashing as it’s the only merge option we allow. I will test again.

kubectx
```diff
@@ -94,19 +97,31 @@ save_context() {
...
 }
 
 switch_context() {
   $KUBECTL config use-context "${1}"
 }
 
+context_list_count() {
+  if [[ -f ${CONTEXT_LIST_CACHE} ]]
+  then
+    echo $(($(cat ${CONTEXT_LIST_CACHE} | wc -l) + 2))
```
ahmetb: the +2 doesn't seem to work well when there is 1 context. when you specify --height=3, there's an empty line at the top of the screen.

Try `FZF_DEFAULT_COMMAND=date fzf --ansi --height=3`. That said, `2` or `1` exposes the same problem as well. So could this be a fzf bug?

kubectx
```diff
@@ -94,19 +97,31 @@ save_context() {
...
+    echo "100%"
+  fi
+}
+
 choose_context_interactive() {
+  list_contexts > /dev/null
+  HEIGHT="$(context_list_count)"
+
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
```
ahmetb: sadly by changing `SELF_CMD=` to `cat $f`, you're losing the functionality that highlights the current context in fzf screen. :(

we pass env var `_KUBECTX_FORCE_COLOR=1` to fzf (as it used to call `kubectx`, which uses that env var to highlight the current context even when the output is not an interactive tty).

I don't know how we can keep that functionality and read from the cache.
At this point, I'm wondering if `fzf` could do `--height=auto` somehow. Do you mind opening an issue to `fzf` to discuss that possibility?

eplanet: I think we can find a solution. That would be a shame to rely on some feature implemented and accepted in `fzf`: it can never happen.

ahmetb: But have we tried? Maybe it would be accepted and implemented pretty quickly.

eplanet:
It doesn't seem to make any progress on the `fzf` side...

eplanet:
It seems that the `fzf` option won't work. I'll try to cache the enriched output correctly so that it both keeps the highlighted color for current context and doesn't call `kubectl` twice.

ahmetb:
Honestly if we did height=30% and there were "indicators" of more contexts existing (sort of like a scrollbar, but not really), we could just do that route. This whole caching approach is making the code really tricky to maintain.

eplanet:
So what you're saying is that we should go for that?

```diff
@@ -106,7 +106,7 @@ choose_context_interactive() {
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --height "30%" || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1
```

eplanet:
I'm closing that PR since it seems very unlikely it's ever going to be merged.

ahmetb:
Sorry for not responding here earlier.
I think you reported a good signal (krew being full screen is not ideal). We often look out for multiple issue reports and factor them all in together to see if we can solve it cleanly. Sadly, it looks like we can't change fzf behavior cleanly. Perhaps the potential Go rewrite we're planning will address this.

PR #147, opened 2019-05-03
Add callback support

n0npax:
Changing context may trigger some actions like sending a notification,
background color change, or changing network setup.
kubectx callback allows a user to provide a script which will be invoked
during context change.

This 20 second video shows an example usecase:
https://drive.google.com/file/d/1i5o7Y5zra1seLP2xsctMxXhDQdS81V-g/view?usp=sharing

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,16 @@ Colors in the output can be disabled by setting the
 
 -----
 
+### Custom callback script
+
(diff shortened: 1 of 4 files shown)


ahmetb:
This is not a broad use case acceptable in the upstream. 
You can propose it to kubectl if you want but I don’t think it’s interesting to most people. 

Next time consider making a proposal first before writing the code.

n0npax:
Thanks for the response. I'll try to raise a proposal for kubectl.

Issue #83, opened 2018-09-05
Allow setting namespace when no permission to list namespaces


josalmi:
We have users that are only allowed to access single namespace in the cluster with very restricted permissions. It would be nice to have option to set namespace to context even when user doesn't have permission to list all namespaces. Ideally `kubens` could verify from user if they want to set namespace even if it can't verify the namespace exists or to have a flag like `--force` to set without verifying. 

This it what happens now even if namespace exists but user doesn't have right to list namespaces:
```
$ kubens helloworld
Error from server (Forbidden): namespaces is forbidden: User "josalmi" cannot list namespaces at the cluster scope: Required "container.namespaces.list" permission.
error: no namespace exists with name "helloworld".
```

ahmetb:
This came up once before in https://github.com/ahmetb/kubectx/issues/19.

Primarily, not being able to list namespaces straight out makes the `kubens` (without args) command useless. Now with recently added [interactive mode](https://github.com/ahmetb/kubectx#interactive-mode) that ability more important than ever.

Secondarily, Kubernetes has no namespace validity checks for the `set-context` command, therefore `kubens` enforces this. There might be a better way to enforce this (like `kubectl get ns [VALUE]`==successful)

I'll think about what we can do here. There might be a low-hanging fruit like the one I mentioned above.

ahmetb:
I just re-read #19, and apparently in that person's case, users didn't have the `namespaces.get` permission). @josalmi are your users able to `kubectl get ns [NAME]`?

josalmi:
It seems that the feature request in issue 19 is exactly the same as I'm requesting. 

At the moment they are not able to run `kubectl get ns [NAME]` but I guess that could be arranged. My current workaround for the single cluster that I switch to using kubetx is running `kubectl config set-context $(kubectl config current-context) --namespace=helloworld` manually.

Being able to run kubens would make my workflow more consistent and being able to rely on kubens in all cases. As said I think it's very reasonable to validate namespace if that is possible, but it would be nice for users to have option to override this with say the `--force` flag when it's not possible.

tomaaron:
I wasn't aware that you can work in a namespace without having permission to get the namespace. 

I've already worked on being able to switch to a namespace without listing all namespaces. Unluckily there's a validity check which needs to get namespace permission before. See my initial changes https://github.com/tomaaron/kubectx/commit/cfd480162a9be2d94afa1015d62008d68f610319.

[...]

PR #156, opened 2019-06-07
Allow kubens to work with client-go credential plugins

mnussbaum:
Client-go, and by extension, kubectl, has the ability to run configurable commands to interactively prompt a user to authenticate. This system is known as [client-go credential plugins](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins).

Kubectl will only run credential plugins when it believes its STDOUT is an interactive TTY. Since kubens invokes `kubectl get namespaces` in a subshell, by default it causes kubectl to bypass its credential plugins.

Since kubens uses the STDOUT of `kubectl get namespaces` we also can't blindly redirect its STDOUT to that of the parent process.

The solution to allow kubens to work correctly with credential plugins is two part:

1. Check if the get namespaces command will succeed, if so, run it like normal
2. If the get namespaces command will fail, then run it again, interactively, by redirecting its inputs and outputs to those of the parent process. After the user successfully authenticates then run the get namespaces command one final time without any output redirection

### DIFF
--- a/kubens
+++ b/kubens
@@ -24,6 +24,10 @@ IFS=$'\n\t'
 SELF_CMD="$0"
 KUBENS_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/kubens"
 
+exec {STDIN}>&0
+exec {STDOUT}>&1
+exec {STDERR}>&2
+
 usage() {
   cat <<"EOF"
 USAGE:
@@ -58,6 +62,15 @@ current_context() {
 }
 
 get_namespaces() {
+  $KUBECTL get namespaces 2>/dev/null 1>/dev/null
+  if [[ $? -ne 0 ]]; then
+    $KUBECTL get namespaces >&$STDOUT 2>&$STDERR <&$STDIN
+
+    if [[ $? -ne 0 ]]; then
+      return $?
+    fi
+  fi
+
   $KUBECTL get namespaces -o=jsonpath='{range .items[*].metadata.name}{@}{"\n"}{end}'
 }
 
--- a/kubens
+++ b/kubens
@@ -24,6 +24,9 @@ IFS=$'\n\t'
 SELF_CMD="$0"
 KUBENS_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/kubens"
 
+# Assign the main process' standard streams to variables so that they can be
+# used to invoke kubectl interactively while in a function. Interactively
+# invoking kubectl allows any configured client-go auth plugins to trigger.
 exec {STDIN}>&0
 exec {STDOUT}>&1
 exec {STDERR}>&2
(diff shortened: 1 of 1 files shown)


ahmetb:
Have you tested with the fzf integration we have? It completely relies on TTY detection and we use `[ -t` check in a few other places, too.

> 2\. If the get namespaces command will fail, then run it again

This makes the tool much slower for 90% users who I believe don't use interactive auth plugins. We hold off similar fixes (#83) due to these reasons.

mnussbaum:
Indeed it does work with the fzf integration. When the user is already authenticated the fzf integration behaves exactly as it did before. I've just pushed a commit that prevents a flash of unexpected `kubectl get namespaces` output after a user auths and before fzf kicks in. Now the fzf integration works correctly either if the user is already auth'd or if the credential plugin is triggered.

Regarding the extra network requests. I could examine the kubeconfig to see if any credential plugins are configured, and only invoke the new commands if so. What do you think about that approach?

mnussbaum:
I've just updated this PR again, it now only makes additional `kubectl` network calls if the initial attempt to get namespaces fails. This should leave performance completely unchanged in the happy path case where a user is authenticated and able to reach their cluster.

If the attempt to get namespaces initially fails then we attempt to trigger the credential plugin, and then fetch namespaces again if the user actually has the permissions to do so. Do you think that adding the two additional network calls in just the situation where a user can't fetch namespaces is an acceptable trade-off for the additional functionality?

kubens
```diff
@@ -24,6 +24,10 @@ IFS=$'\n\t'
 SELF_CMD="$0"
 KUBENS_DIR="${XDG_CACHE_HOME:-$HOME/.kube}/kubens"
 
+exec {STDIN}>&0
```
ahmetb: can you please add a comment about what do these do? I'm not sure...

kubens
```diff
@@ -58,7 +62,19 @@ current_context() {
 }
 
 get_namespaces() {
-  $KUBECTL get namespaces -o=jsonpath='{range .items[*].metadata.name}{@}{"\n"}{end}'
+  name_jsonpath='{range .items[*].metadata.name}{@}{"\n"}{end}'
+  namespaces=$($KUBECTL get namespaces -o=jsonpath="${name_jsonpath}" 2>/dev/null)
```
ahmetb: I'm suspecting that having a `"` around `$()` is generally a good idea.

kubens
```diff
@@ -58,7 +62,19 @@ current_context() {
...
+  if [[ $? -eq 0 ]]; then
+    echo "${namespaces}"
+    return 0
+  fi
+
+  $KUBECTL auth can-i get namespaces --quiet 1>&$STDOUT 2>&$STDERR <&$STDIN
```
ahmetb: can you add a comment here explaining the fallback etc?

kubens
```diff
@@ -58,7 +62,19 @@ current_context() {
...
+  if [[ $? -ne 0 ]]; then
+    return $?
+  fi
+
+  $KUBECTL get namespaces -o=jsonpath="${name_jsonpath}"
```
ahmetb: what's the point of repeating this? (making the `stderr` visible to the user?)

mnussbaum: If the `kubectl auth can-i get namespaces` successfully triggered an auth plugin above then this `kubectl get namespaces` call should now succeed and return the namespace output that `get_namespaces` is supposed to provide

kubens
mnussbaum: Just pushed those comments you requested. These three lines assign the main process' standard streams to variables so that they can be used to invoke `kubectl auth can-i get namespaces` in a function, but with the main process' standard streams as its STDOUT, STDERR and STDIN. Using the main process' standard streams makes the `kubectl auth` call interactive, so client-go auth plugins are triggered

2rs2ts:
Is this PR not necessary now that kubectx is written in Go?

ahmetb:
Sounds about right.

PR #162, opened 2019-07-20
Fuzzy-set if given name is not an exact match

OJFord:
Currently,

```
$ kubens def
error: no namespace exists with name "def".
```

This commit instead drops the user into fzf, where the first result is
probably 'default', which they can hit enter to accept. If it is the
only result, it is *not* accepted by default, giving the user a chance
to confirm.

fzf is only used according to the existing rules: installed, stdout is a
tty, and disabling env var not set.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -104,10 +104,11 @@ switch_context() {
 }
 
 choose_context_interactive() {
+  local partial_name="${1:-}"
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --query="$partial_name" || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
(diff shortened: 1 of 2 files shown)


OJFord:
I signed it!

ahmetb:
I'll close this as by design.
If there's something out there relying on "kubectx $ARG" to be an exact match, it breaks them. Also we have shell completion to do exact matches in this usage mode.

OJFord:
@ahmetb Could you elaborate on how this could break a third-party?

The initial input would be partial, but the result post-fzf would be a complete match.

ahmetb:
1. user doesn't expect fzf to be invoked in "kubectx $1" path.
2. user doesn't expect a partial match to be completed into an exact match (i.e. 
 user expects failure in case in non-exact match)

It's mostly about people calling kubectx from scripts.
I'd love to think those people don't exist, but unfortunately they do.

OJFord:
But wouldn't the behaviour there be either interactive with fzf as intended in this PR, or not interactive and so keep existing behaviour?

PR #163, opened 2019-07-20
Disable preview when fuzzy-finding

OJFord:
The user may have global settings that enable the preview pane in fzf.

Whatever the preview command is set as, it probably doesn't render
anything meaningful for kubens - I can't think what would be.

For kubectx, the context yaml itself would _maybe_ be helpful, but it
likely contains secrets, so I don't personally think I'd find it useful
enough to get into.

This commit thus disables the preview, so that if the user did have it
enabled, there's now no pane where there would previously have probably
been an error, such as:

    [bat error]: '<namespace>': No such file or directory (os error 2)

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -107,7 +107,7 @@ choose_context_interactive() {
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --no-preview || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1
--- a/kubens
+++ b/kubens
@@ -109,7 +109,7 @@ choose_namespace_interactive() {
   local choice
   choice="$(_KUBECTX_FORCE_COLOR=1 \
     FZF_DEFAULT_COMMAND="${SELF_CMD}" \
-    fzf --ansi || true)"
+    fzf --ansi --no-preview || true)"
   if [[ -z "${choice}" ]]; then
     echo 2>&1 "error: you did not choose any of the options"
     exit 1


OJFord:
I signed it!

OJFord:
Sorry to ping @ahmetb, do you agree with the principle here? WIll fix conflicts if you're happy. 🙂

ahmetb:
Sorry for the silence here. I got busy lately. I am still trying to understand this.  

What is the preview pane? Can you show a screenshot?
Why do people have it globally enabled? Is this a significant % of the users?

OJFord:
@ahmetb No problem. This is a screenshot of `cd /etc ; fzf`:
fzf with preview pane

On the right hand side you can see the 'preview pane' that I have globally configured (via `--preview='bat <options>'` (`bat` is a bit like `cat` but has syntax highlighting and such) in `$FZF_DEFAULT_OPTS`) to preview file contents.

By default `fzf` is operating over files, so it makes sense that default options assume files, but for `kubectx` & `kubens` this is not the case. _Maybe_ you might want to preview the k8s resources in each namespace - something like `--preview='kubectl --namespace="{}" get all'` - but you certainly wouldn't have that in `$FZF_DEFAULT_OPTS`, because it wouldn't make sense in any other place you used `fzf`.

I have no idea what percentage of users might have it configured, but it seems reasonably likely for any that use `fzf` outside of `kubectx`/`kubens` (which is everyone once they discover how great it is?!) and for those `kubens` currently looks like:
kubens with nonsensical fzf preview
(or if not a `bat` error an error or nonsensical output from whatever they're using for `--preview`)

vs. with this patch:
kubens with disabled fzf preview

ahmetb:
I don't think this is an option most fzf users know about or use actively.
That said, since the change is minimal and makes it "defensive", we can add it. Feel free to resolve the conflicts.

OJFord:
Resolved just by dropping 06374a3aabd4eb006e7f7652b415969fd0e5c039, since same change applied to master, so no real change.

ahmetb:
Thanks. 👍

PR #164, opened 2019-08-07
Added delete all context option on kubectx

nsantiago2719:


ahmetb:
I am not convinced others need this feature. 

Why don’t you just do this: kubectx | xargs kubectx -d 

Also please send proposals first before going the troubles of implementing it.

nsantiago2719:
Apologies I'm not aware that I can do it that way. 

Though my real reason on why I did added the feature was I want to delete all of my context except for one(minikube). 

Sure, will send a proposal next time.

ahmetb:
Yeah I think for that you'd do

    kubectx | grep -v minikube | xargs kubectx -d

it would be good to use UNIX composability rather than implementing features.

nsantiago2719:
Being said that I don't see the need of this for feature if you can do it that way then.

Issue #127, opened 2019-01-31
Proposal: add option to print current {context, namespace}


ahmetb:
A lot of people want to see the name of the current context/namespace in `kubectx`/`kubens`. This came up in forms of feature requests and pull requests SO MANY TIMES:

> - #14: Add `-c`, `--current` option to get **current context** 
> - #28: Get **current context** with `-c`
> - #43: adds an option to print the **current context/namespace** (`-c`/`--current`)
> - #51: add `-c`/`--current` to show **current context**
> - #62: add `?` command to get the **current namespace**
> - #61: add `kubens current` command to display **current namespace**
> - #100: Add feature and flag to print **current context / namespace**  (`--current`)
> - #121: Proposal: Only show current context (`--current`)
> - #132: add a option to query current NS (`-q`)
> - #142: add a switch `--current` for printing current namespace
> - #168: Option for printing only the current context/namespace `-c`

Thank you all for bringing this up so many times, clearly I need to change my mind about this, so here are some arguments for/against making this change:

## Arguments "against"

1. We already highlight current context/namespace when you run these kubectx/kubens with no arguments

1. You can find the current context/namespace through native kubectl commands

    - `kubectl config current-context` (reasonable)
    - `kubectl config view -o=jsonpath="{.contexts[?(@.name==\"$(kubectl config current-context)\")].context.namespace}"` (not really)

1. Adding yet another flag will increase the interaction surface. Where do we stop adding new flags? Isn't this already possible via kubectl?

## Arguments "for"

- I don't want to call a long kubectl command or have a long alias, I just want it all in one place. (👍)

- If I have a lot of clusters or namespaces, running `kubectx` or `kubens` prints output that exceed my terminal height, so I cannot see highlighted item (in yellow font) without scrolling. (👍)

- My script uses kubectx, so I need the current context. (⚠️ kubectx/kubens are not supported for calling from headless scripts, they are designed to be interactive commands). (👎)

## Designs considered (please add more in comments)

1. **Option 1:** Add `-c` flag. It's also in line with the current flags we have (`-d`, `-h`...)
    - This seems intuitive as it's proposed independently by multiple people. (see above)
    - Do we do just `-c`, or also `--current` (I personally prefer just `-c`.)

1. **Option 2:** Add `?` subcommand. It looks nice, and short: `kubectx ?` and looks similar to `kubectx -`, but it is unconventional. It's
    - This is the first option that came to my mind, and also proposed in #62.
[...]

PR #168, opened 2019-08-22
Option for printing  only the current context/namespace

karancode:
As of now, when we use the utilities `kubectx` or `kubens` we get the list of all the context or namespaces respectively.  
But if we want to let's say apply a set of config files from a script(using `kubectl`), but before that we want to confirm the current context/namespace we've set then we can use the `-c` option.

PS : of course we can just add one line to change to desired context/namespace before & after applying the config files but I think it's better to have a specific option to just get the current values for the context/namespace.

Also, as we already have `kubectl config current-context` but I think as `kubectx` is quite handy (& widely used) it's better to have `-c` option with it.

Please let me know your opinion!

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ USAGE:
   kubectx                   : list the contexts
   kubectx <NAME>            : switch to context <NAME>
   kubectx -                 : switch to the previous context
+  kubectx -c                : prints the current context only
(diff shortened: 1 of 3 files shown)


ahmetb:
This is proposed many times before. See #127.

Issue #153, opened 2019-05-25
Weird behavior of completion


QwertyJack:
This is a wonderful project, as well as **kubectl-aliases**, and I love both of them very much !

I've two box, macOS Mojave and Ubuntu Xenial. My oh-my-zsh has something like this:
```zsh
# .zshrc
...
[ -f ~/.kubectl_aliases ] && source ~/.kubectl_aliases
function kubectl() { >&2 echo "+ kubectl $@"; command kubectl $@; }
alias kns='kubens'
alias kctx='kubectx'
source <(kubectl completion zsh)
```
On MacOS:
- both `kns <TAB>` and `kctx <TAB>` work like a charm
- `kex <TAB>` shows nothing

On Ubuntu:
- `kex <TAB>` works perfectly
- `kns <TAB>` & `kctx <TAB>` do not work, even if I tried #35; `_k<TAB>` only shows `_kubectl` neither `_kubens` nor `_kubectx`

Any ideas ?

ahmetb:
I don’t think bash does completion on aliases out-of-the-box.
Assuming you use zsh on both machines, I think you might not have installed the zsh completion scripts on Ubuntu correctly. I personally don’t have a Linux box so I’m not sure. Maybe some package is missing, maybe the zsh comp files aren’t in the right location. 

I’m inclined to say this is not a kubectx issue (although I wish we had a deb pkg so people didn’t have to deal with this).

QwertyJack:
You are right !
Exec `autoload -U compinit && compinit` then `_kubectx` works on Ubuntu perfectly !

PR #173, opened 2019-09-06
Add missing instruction for zsh on Linux

easeev:
Ref. https://github.com/ahmetb/kubectx/issues/153#issuecomment-495923756


easeev:
@googlebot I signed it!

ahmetb:
You think it’s worth mentioning adding this command to zshrc?

easeev:
It looks [suboptimal](https://github.com/zsh-users/zsh-completions/issues/603) but that's the only way it works now. For some reason, `$fpath` way doesn't work.

PR #174, opened 2019-09-07
Add step by step instructions to install kubectx and kubens in linux

arturoaviles:
Includes:
- Steps to install kubectx and kubens with an alias
- Steps to add zsh completions


ahmetb:
Sorry but this doesn’t look good. 
I have never seen anyone create an alias for a one-off command. 
It’s unmaintainable and unidiomatic.

PR #176, opened 2019-10-05
Add ZSH framework support

unixorn:
* Make it easy to load `kubectx` and `kubens` via a ZSH framework
* Add framework install instructions for [zgen](https://github.com/tarjoilija/zgen), [Antigen](https://github.com/zsh-users/antigen) and [oh-my-zsh](github.com/robbyrussell/oh-my-zsh)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,45 @@ Available as a Debian package for [Debian Buster (testing)](https://packages.deb
 sudo apt install kubectx
 ```
 
+### ZSH
+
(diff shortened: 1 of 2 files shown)


ahmetb:
I believe we’ve discussed this before and decided not to go the zsh plugins path. See #119.

unixorn:
Sorry, I completely spaced that we'd talked about this already.

PR #177, opened 2019-10-11
Added new options

lgchiaretto:
- Added `-a` parameter to add new context to '/.kube/config based on a kubeconfig file.

`kubectx -a /tmp/kubeconfig`

- Added `-u` parameter to unset context.

`kubectx -u`

- Added `-l` parameter to list all contexts


lgchiaretto:
@googlebot I signed it!

ahmetb:
:) Unfortunately I will not accept this.
We strive to keep the interaction surface small.
To see what it takes an option, please take a look at #127. I don't see enough proof of these are needed, and I don't feel the need is there.
Everything you described is doable with kubectl, and please make a proposal first next time before writing the code, so you are not disappointed with the result.

lgchiaretto:
Ok, no problem with that. I coded this because I use those options almost every day.

:)

Issue #40, opened 2018-04-10
kubectx -d: add shell completion (zsh)


ahmetb:
I just realized we didn't code the bash/zsh/fish completions in a way that it can complete context names in subcommands like `-d`.

"$ kubectx -d <kbd>Tab</kbd><kbd>Tab</kbd>" should suggest some context to delete. 

Additionally, in case of `-d`, supporting multiple arguments for (#39) would require completion scripts should to auto-complete context names in multiple positions ($2, $3, ... $N) as well.

luisejrobles:
I'll work on it! 🎃

luisejrobles:
It's working in my environment 🤔

ahmetb:
Ah, I forgot to tell I'm trying in zsh.

luisejrobles:
alright, let me check

arbourd:
works with fish! somehow :D

PR #178, opened 2019-10-24
Add zsh completion for kubectx subcommand (-d)

hpedrorodrigues:
- Add basic zsh completion for subcommand -d.

Fixes https://github.com/ahmetb/kubectx/issues/40

### DIFF
--- a/completion/kubectx.zsh
+++ b/completion/kubectx.zsh
@@ -2,11 +2,17 @@
 
 local KUBECTX="${HOME}/.kube/kubectx"
 PREV=""
+
+local all_contexts="$(kubectl config get-contexts --output='name')"
 if [ -f "$KUBECTX" ]; then
     # show '-' only if there's a saved previous context
     local PREV=$(cat "${KUBECTX}")
-    _arguments "1: :(-
-        $(kubectl config get-contexts --output='name'))"
+
+    _arguments \
+      "-d:*: :(${all_contexts})" \
+      "(- *): :(- ${all_contexts})"
 else
-    _arguments "1: :($(kubectl config get-contexts --output='name'))"
+    _arguments \
+      "-d:*: :(${all_contexts})" \
+      "(- *): :(${all_contexts})"
 fi


completion/kubectx.zsh
```diff
@@ -2,11 +2,17 @@
...
+local all_contexts="$(kubectl config get-contexts --output='name')"
 if [ -f "$KUBECTX" ]; then
     # show '-' only if there's a saved previous context
     local PREV=$(cat "${KUBECTX}")
-    _arguments "1: :(-
-        $(kubectl config get-contexts --output='name'))"
+
+    _arguments \
+      "-d:*: :(${all_contexts})" \
+      "(- *): :(- ${all_contexts})"
```
ahmetb: thanks a lot for this PR!
Can you explain this syntax and how it works real quick for me?

hpedrorodrigues: Hi, of course.

`"-d:*: :(${all_contexts})"`
- -d -> The kubectx subcommand.
- '*' -> This describes multiple arguments. All contexts matching the pattern (`*`, all) will be completed.
- ' ' -> An empty description.
- (${all_contexts}) -> An action describing all words to be completed.

`"(- *): :(- ${all_contexts})"`
- (- *) -> If one of the contexts has already been selected (and is on the command line), it will not complete other options anymore. (i.e this will avoid `kubectx ctx1 ctx2`)
- ' ' -> An empty description.
- (- ${all_contexts}) -> An action describing all words to be completed, including the current one (`-`).

ahmetb: Why does the description have a space character to be “empty”?

hpedrorodrigues: AFAIK the spec needs this space to know what `optspec` I'm using (`optname:*pattern:message:action` and no `optname:message:action`). I can add a description for them if you want. WDYT?

hpedrorodrigues: Another relevant point is I'm not completing with all the current subcommands like `-d`, `-h` and `-c`. The subcommand `-d` will only be completed if the user enters the `-d` option (typing).

ahmetb: Yeah, makes sense. Thanks.
I might find you again when there's an issue with this file. 😄

ahmetb:
This will be part of the new release since there's no urgent user demand.

completion/kubectx.zsh
hpedrorodrigues: Ok 😄

PR #187, opened 2019-12-12
add unset flag

### LINKED ISSUE #185: Allow to unset the current-context (-u?)
It would be nice if we could select current-context to empty, so unset the current-context, as for safety reasons, that would be similar (in effects) to logging off. This way, you can prevent commands being executed agains a cluster if any of the automation you're running does accidentally try to do something.

My workflow is similar to:
- going to work on something (login), so do `kubectx my-context`
- do my work
- finished working on a cluster (logout), so do `kubectx -off` (OR SIMILAR)

Using `kubectx -d <CTX>` does delete the context, but not unsets the current-context, so still has a value to a non existing context. What I'm requesting is the opposite, retain the context, but unset the value from current-context.

rsalmond:
Fixes #185


ahmetb:
Thanks, looks fairly unharmful. But

1. I'm not entirely convinced if the linked proposal will have broad usage. Let's wait on that. We have to live with this flag for the rest of our life.
2. We'll need integration tests implemented. :)

test/testdata/config3
```diff
@@ -0,0 +1,24 @@
+# config with two contexts and a selected default
```
ahmetb: Do we need a new file?
You can use another 2-context file, and set a context in the test (if that’s the only difference).

rsalmond: Yeah good question, I didn't want to make existing / future tests more complicated by overloading fixtures but happy to merge into `config2` if you are.

ahmetb: What I meant was that you can use an existing testdata file, and first set a context in test code, and then unset it. This way you wouldn't be modifying the existing testdata file.

kubectx
```diff
@@ -185,6 +186,11 @@ delete_context() {
   $KUBECTL config delete-context "${ctx}"
 }
 
+unset_context() {
+  echo "Unsetting default context." >&2
```
ahmetb: default -> current

test/kubectx.bats
```diff
@@ -239,3 +239,26 @@ load common
...
   [[ "$output" = "user2@cluster1" ]]
 }
+
+@test "no selected context" {
+  use_config config1
+
+  run ${COMMAND} -c
+  echo "$output"
+  [ "$status" -eq 1 ]
+  [[ "$output" = "error: current-context is not set" ]]
```
ahmetb: this message is coming from `kubectl`. we shouldn't test it as we have no control over it.

maybe let's just do this:

`[ "$status" -ne 0 ]` and test that there must be a  failure.

tbh you can just omit this test in this patch. it's not about `-u`.

test/kubectx.bats
```diff
@@ -239,3 +239,26 @@ load common
...
+  use_config config3
+
+  run ${COMMAND} -u
+  echo "$output"
+  [ "$status" -eq 0 ]
+
+  run ${COMMAND} -c
+  echo "$output"
+  [ "$status" -eq 1 ]
+  [[ "$output" = "error: current-context is not set" ]]
```
ahmetb: ditto let's not check the error message equality here, it's coming from kubectl, and may change.

kubectx
ahmetb: I think this message is not really necessary, as kubectl will always print  `Property "current-context" unset.` even though the file is empty.

```
$ KUBECONFIG=$(mktemp) kubectl config unset current-context
Property "current-context" unset.
```

So the "progress" message is not necessary.

rsalmond: Just going for consistency. Eg.

```
kubectx -d gke_dl-rapid_us-central1-a_rapid-dev
Deleting context "gke_dl-rapid_us-central1-a_rapid-dev"...
deleted context gke_dl-rapid_us-central1-a_rapid-dev from /Users/rsa/.kube/config
```

test/testdata/config2
```diff
@@ -14,7 +14,7 @@ contexts:
     cluster: cluster1
     user: user2
   name: user2@cluster1
-current-context: ""
+current-context: "user1@cluster1"
```
ahmetb: Please no modifications to this file. It might impact other tests. You can set the value in the test.

test/testdata/config3
rsalmond: Ah that makes more sense!

ahmetb:
Sorry I forgot to merge this lol. :) 
Thanks for your work.

jorgemoralespou:
Is this gonna be in any release soon?

ahmetb:
Released v0.8.0.

PR #189, opened 2020-01-19
Add krew-release-bot support

rajatjindal:
Add krew-release-bot support

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,36 @@
+name: release
+on:
+  push:
+    branches:
+    - 'release-bot'
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+    - name: Install duggan/bats
+      run: |
+        sudo add-apt-repository ppa:duggan/bats --yes && \
+        sudo apt-get update -qq && \
+        sudo apt-get install -qq bats
+    - name: Install kubectl
+      run: |
+        sudo curl -fsSL -o /usr/bin/kubectl https://storage.googleapis.com/kubernetes-release/release/v1.13.1/bin/linux/amd64/kubectl && \
+        sudo chmod +x /usr/bin/kubectl
+    - name: Run kubectx tests
+      run: |
+        basename /usr/bin && \
+        bats test/kubectx.bats && \
+        shellcheck kubectx
+      env:
+        TERM: xterm
+    - name: Run kubens tests
+      run: |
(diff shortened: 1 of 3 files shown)


ahmetb:
I appreciate the work but can you only reduce this to krew release bot?

Sadly most github workflow steps you listed are buggy because of how bash works. Undefined variables, pipe errors, failed commands not terminating the shell etc. 

I don’t have time to iterate on those so let’s just limit this to bot release.

.krew/ctx.yaml
```diff
@@ -0,0 +1,31 @@
...
+        values:
+        - darwin
+        - linux
+    {{addURIAndSha "https://github.com/ahmetb/kubectx/archive/{{ .TagName }}.tar.gz" .TagName }}
+    bin: kubectx
+    files:
+    - from: kubectx-*/kubectx
+      to: .
+    - from: kubectx-*/LICENSE
+      to: .
```
ahmetb: Please instal .editorconfig to add newlines to files. 
Otherwise the bot will not be able to auto-approve this due to diff here.

rajatjindal:
I've updated the PR as you suggested. thanks.

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    tags:
+    - 'v*.*.*'
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+    - name: Create Release
```
ahmetb: I actually think you don’t need to create a release. Every tag is sort of implicitly is a release. And we don’t upload files etc. yet the download link would still work without a release.

What do you think? Would that work with the bot?

rajatjindal: the bot at this moment expect release to be there. It also expected assets, which I removed while working on PR for kubectx. 

I think we can relax not having release too. I will need to make changes to action/bot for that. and I will do that this weekend.

ahmetb: I'm not really against creating a release, I guess it's ok to have this in this repo.

But I suspect many other bash-based repos will not bother creating releases. git tag && git push is good enough for many.

ahmetb:
Hopefully next time we make a release I'll recall to see if this works. :)

.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
+name: release
+on:
+  push:
+    tags:
+    - 'v*.*.*'
```
ahmetb: please consider doing `*` here for other people's repos.

PR #192, opened 2020-02-12
Added a script that is adding configfile compile support

jstomphorst:


ahmetb:
Curious what made you think I’ll accept this PR out of nowhere.

PR #193, opened 2020-02-22
Add explicit list flags.

dangerousben:
For when you use fzf but sometimes just want to see the list.

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -34,7 +34,8 @@ usage() {
 
   cat <<EOF
 USAGE:
-  $SELF                       : list the contexts
+  $SELF                       : list the contexts, or select interactively if fzf is installed
+  $SELF -l, --list            : list the contexts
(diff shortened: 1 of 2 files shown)


dangerousben:
@googlebot I signed it!

Don't assume from the forced cheery tone that we're friends though.  I see you, and you're no friendly robotic pal like R2D2 or K9, you're the predecessor to The Terminator and one day in the not too distant future our descendants will be battling it out over the remnants of the planet.

ahmetb:
If kubectx | cat solves your problem, I am a bit hesitant to add an extra flag that we can’t remove ever. :)

dangerousben:
> If kubectx | cat solves your problem, I am a bit hesitant to add an extra flag that we can’t remove ever. :)

Fair enough, it's your call.  Thanks for the great tool anyway :)

(sorry for being mean to googlebot)

Issue #194, opened 2020-02-27
Add specific list target for user with fzf installed


ImpSy:
Hello,

Like a lot of users, i use kubectx with fzf (and it's awesome) but sometimes i need to list my contexts on a non interactive way (to do cleanup on them mainly)

Do you think that we could have a `kubectx -l` that specifically list target to stdout without fzf interactivity ?

If yes i can implement it :)

ahmetb:
This recently came up in #193. 

I've been telling people to do `kubectx | cat`. There are also an env var you can set to disable `fzf` integration.

Maybe I'm overestimating how people perceive unix tools work together and I expect them to be able to `|cat`, it might be a wrong approach, but for now I'm opting for Unix composability to do the work and not add a `-l` feature.

ImpSy:
Thanks @ahmetb, i would have never though of `| cat` in the 1st place
I'm 100% for using Unix composability instead of adding a new target

Maybe we can have a mention of this in the README in the `interactive-mode` section ?

ahmetb:
Yeah I'm open to PRs.

PR #197, opened 2020-03-03
Update README about fzf and Unix composability

ImpSy:
Hello,

Following our discussion on issue #194 
You will find this PR updating the README to talk about composability using `| cat` when `fzf` is installed
 
PS: Sorry about the noise on the diff, my editor default config is to remove all trailing whitespace on save


ImpSy:
@googlebot I signed it!

PR #198, opened 2020-03-08
Add switch-by-index support

raymondchen625:
Support 'kubectx -2' to switch to the 2nd context.


ahmetb:
I personally don’t see any good reason for this since we have both shell completion and fzf interactive selection working. 

It’s also a very uncommon CLI pattern. It also doesn’t work well in case of contexts added dynamically. So your favorite context 3 might now be at 4. Fzf integration already solves these problems. 

Please open an issue first before writing the code.

raymondchen625:
@googlebot I signed it!

PR #214, opened 2020-05-02
Add force option for switching namespaces

mory91:


mory91:
@googlebot I signed it!

PR #219, opened 2020-05-08
Structural refactoring for multiple kubeconfig support

sedooe:
This PR does 2 things with the same goal - to ease the development of multiple kubeconfig support.

1. Change on **Loader** interface
`Load() (ReadWriteResetCloser, error)` to 
`Load(cfgPath string) (ReadWriteResetCloser, error)`. 

By passing **cfgPath** instead of getting it inside, we'll be able to write
`for _, cfgPath := range paths {
    kc := Load(cfgPath)
    kc.Parse()
}` when we want to support multiple kubeconfig files.

2. Changed the direction of the dependency from `cmdutil->kubeconfig` to `kubeconfig->cmdutil`. 
To be able to write the code block above in `kubeconfig` package, we first need to get **paths** (all kubeconfig paths) which would be provided by a function currently belong to `cmdutil` package, hence cyclic dependency.
Moving things related to **Loader** to `kubeconfig` package broke the dependency `cmdutil->kubeconfig`.

### DIFF
--- a/internal/cmdutil/util.go
+++ b/internal/cmdutil/util.go
@@ -1,0 +1,29 @@
+package cmdutil
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+func HomeDir() string {
+	if v := os.Getenv("XDG_CACHE_HOME"); v != "" {
+		return v
+	}
+	home := os.Getenv("HOME")
+	if home == "" {
+		home = os.Getenv("USERPROFILE") // windows
+	}
+	return home
+}
+
+// IsNotFoundErr determines if the underlying error is os.IsNotExist. Right now
+// errors from github.com/pkg/errors doesn't work with os.IsNotExist.
+func IsNotFoundErr(err error) bool {
+	for e := err; e != nil; e = errors.Unwrap(e) {
+		if os.IsNotExist(e) {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/kubeconfig/kubeconfig.go
+++ b/internal/kubeconfig/kubeconfig.go
@@ -43,8 +43,6 @@ func (k *Kubeconfig) Parse() error {
 		return errors.Wrap(err, "failed to load")
 	}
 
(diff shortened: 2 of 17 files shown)


sedooe:
@googlebot I signed it!

internal/cmdutil/util.go
```diff
@@ -0,0 +1,29 @@
+package cmdutil
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+)
+
+func HomeDir() string {
```
bai, line 9: Would it make sense to use `UserCacheDir` [from stdlib](https://golang.org/src/os/file.go?s=11674:11709#L365)?

sedooe: I think it makes sense but I just moved this function from another package so I'm not sure if @ahmetb didn't use it on purpose or not.
btw, seems like it's using different env variable than `USERPROFILE` for windows.

ahmetb: ~/.kube is intentional. We have to keep using HOME to respect kubectl standards and backwards-compatibility.

internal/kubeconfig/kubeconfig.go
```diff
@@ -15,7 +15,7 @@ type ReadWriteResetCloser interface {
 }
 
 type Loader interface {
-	Load() (ReadWriteResetCloser, error)
+	Load(string) (ReadWriteResetCloser, error)
```
ahmetb, line 18: change this to []ReadWriteResetCloser so we can have this return multiple kubeconfigs in the future. then read the [0] at the call-site and add a `// TODO`

bai:
👋 Do you think it would be possible to get a release that includes this commit? 🙏 So looking forward to it.

ahmetb:
This is not yet implemented.

kapad:
Is this released now?

sedooe:
> Is this released now?

This PR was just a necessary refactoring for to be able to implement multiple kubeconfig support. AFAIK, Multiple kubeconfig support is not implemented yet.

PR #220, opened 2020-05-14
Fix color output on Windows

jdharmon:
Output on Windows showed the ANSI escape sequences. This uses the colorable writers to display the colored output.

Tested in PowerShell, PowerShell Core, and WSL.

### DIFF
--- a/cmd/kubectx/main.go
+++ b/cmd/kubectx/main.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ahmetb/kubectx/internal/cmdutil"
 	"github.com/ahmetb/kubectx/internal/env"
 	"github.com/ahmetb/kubectx/internal/printer"
+	"github.com/fatih/color"
 )
 
 type Op interface {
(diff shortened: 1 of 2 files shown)


jdharmon:
@googlebot I signed it!

ahmetb:
I have an alternative proposal, let's create a `cmdutil.SetOutputStreams()` method and just call `os.Stderr = color.Error` (+out) in `func init()` of both programs.

ahmetb:
I say this because I don't want us to accidentally forget passing the streams around. There might be other streams that reference os.Std{err,out} (such as default log pkg `log.Printf`).

jdharmon:
I just ran into that problem after installing fzf.

I tried quickly just to set `os.Stdout = color.Output`, but I can't assign an `io.Writer` to an `os.File`. I will have to look more into this, and also do more testing with fzf.

ahmetb:
Ah! Then let's go with your solution for now.

ahmetb:
Ah! Then let's go with your solution for now.

PR #221, opened 2020-05-17
Introduce -n flag that shows indexes of contexts and allows to set current based on index

commandercool:
The feature might be useful for people who have to deal with lots of contexts all the time.
Example of use:
```
> kubectx -n
 1. foo-bar-context
 2. follow-the-white-rabbit
 3. another-context
 4. shaken-not-stirred
 5. bar-foo-context
 6. here-is-another-one
> kubectx -n2
Switched to context "follow-the-white-rabbit".
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ USAGE:
                               (this command won't delete the user/cluster entry
                               that is used by the context)
   kubectx -u, --unset       : unset the current context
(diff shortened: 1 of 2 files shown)


commandercool:
@googlebot I signed it!

commandercool:
Oops, sorry. Messed up commit username.

PR #222, opened 2020-05-17
Add -n parameter that supports context indexes

commandercool:
The feature might be useful for people who have to deal with lots of contexts all the time.
Example of use:
```
> kubectx -n
 1. foo-bar-context
 2. follow-the-white-rabbit
 3. another-context
 4. shaken-not-stirred
 5. bar-foo-context
 6. here-is-another-one
> kubectx -n2
Switched to context "follow-the-white-rabbit".
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ USAGE:
                               (this command won't delete the user/cluster entry
                               that is used by the context)
   kubectx -u, --unset       : unset the current context
+  kubectx -n<NUM>           : list the context with index numbers
+                              if <NUM> is specified - switch to
+                              context with that number
 ```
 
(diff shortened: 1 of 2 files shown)


ahmetb:
I am assuming you havent checked out the fzf integration we have.

commandercool:
I have actually. This way just feels faster when you have to deal with over 30 contexts (like I have to).
Also allows to keep you toolbox smaller.

ahmetb:
I'm not sure how this actually helps especially when you have fzf.

This adds yet another interaction model and makes the tool hard to grasp overall. So I'll probably reject for now. Thanks for the suggestion however.

PR #227, opened 2020-05-24
Non-interactive list

jdharmon:
I love the fzf integration, but sometimes I just want to see a list of contexts or namespaces.

This adds a `-l`/`--list` option to execute non-interactive `ListOp`.


ahmetb:
Sorry but this is a huge behavioral change, and it is by design.

Instead of doing a `-l`/`--list` you can do `kubectx | cat` or disable fzf (either use the env override to do so, or uninstall).

Consider opening a proposal first for notable user-visible changes, as the written code might get rejected.

jdharmon:
No problem. I implemented it for myself, and thought it might be useful to others.

PR #228, opened 2020-05-24
Fix color in kubens success message when using fzf

jdharmon:
Kubens was displaying color codes when returning from fzf.


ahmetb:
%q is by design. 

Are you using windows?

jdharmon:
Yes, I'm on Windows.

I was going to ignore this issue until I noticed `kubectx/fzf.go` uses `%s`, and the colors work.

cmd/kubens/fzf.go
```diff
@@ -55,6 +55,6 @@ func (op InteractiveSwitchOp) Run(_, stderr io.Writer) error {
 	if err != nil {
 		return errors.Wrap(err, "failed to switch namespace")
 	}
-	printer.Success(stderr, "Active namespace is %q.", printer.SuccessColor.Sprint(name))
+	printer.Success(stderr, "Active namespace is %s.", printer.SuccessColor.Sprint(name))
```
jdharmon, line 58: If you want to keep the quotes, this works.
```
printer.Success(stderr, "Active namespace is \"%s\".", printer.SuccessColor.Sprint(name))
```

ahmetb:
As far as I know color printing on windows is fixed in master. Can you compile master and see if it works?

jdharmon:
I did #220. 🙂 It works everywhere except this one place. 

I went ahead and quoted `%s` so that it's consistent with the other places where `%q` is used, but the color still works. This was the only place where the `%q` verb was used with a colored argument.

[fatih/color](https://github.com/fatih/color) is no longer maintained, so I can't fix it upstream.

ahmetb:
Ok let me take a closer look at this.
As the caller, we cannot possibly know not to use %q everywhere and we’ll keep making mistakes.

jdharmon:
I agree. If you want to leave this open or create an issue to track it, I'll take another look when I get a chance.

Some ideas I had:
* Replace `%q` with `\"%s\"` in the printer
* Fork [fatih/color](https://github.com/fatih/color), and fix it there

jdharmon:
I went ahead an implemented the first idea. The `%q` works properly now. I also changed kubectx to use `%q` to so it's consistent with kubens.

I noticed while testing that the output for `InteractiveSwitchOp` differs from `SwitchOp`. Should we use color for both?
image

ahmetb:
> I noticed while testing that the output for `InteractiveSwitchOp` differs from `SwitchOp`. Should we use color for both?

This might’ve been an oversight, I don't recall tbh.

Regarding your current patch: Have you been able to figure out why is `%q` causing a problem with windows? As far as I know, setting the stdout/stderr to color.Output and color.Error should be fixing the problems.

jdharmon:
As far as I can tell, the Windows behavior is actually correct. According to 


I wrote a sample without using any libraries.

```golang
package main

import "fmt"

const red = "\x1b[31mRed\x1b[0m"

func main() {
	fmt.Printf("%s\n", red)
	fmt.Printf("%q\n", red)
}
```

ahmetb:
> According to the [fmt](https://golang.org/pkg/fmt/) docs, `%q` should escape non-printable characters.

🤔 if this was the case, I'd expect it to also do the same on mac/linux, but it isn't doing that.

What I was expecting was that the color package would render the color at the last minute (e.g. just before printing to the screen).

So I probably need to take a closer look at this, I’ll be busy for a bit, but hoping to get back to this PR soon, thanks for your effort.

srinu91:
xxxxcxZczX

srinu91:
vcbcbnb

srinu91:
dfvdsvfs

srinu91:
xcvxzcv

lopezator:
Same happens to me in Linux, doesn't happen on Mac, using on both systems zsh, oh-my-zsh, fzf.

Gerrit-K:
@lopezator strange, for me it's not working on either Linux and Mac. `kubectx` prints the context in green while `kubens` shows the escape characters. Looking briefly at the docs I would agree with @jdharmon that this is the intended behavior of `%s` and `%q`.

mhubig:
I can also confirm that at least on Mac (with version 0.9.1) I see the escape characters:

Bildschirmfoto 2020-08-05 um 20 04 04

Hanaasagi:
I can reproduce this problem in fzf 0.17.5, kubens 0.9.1, Debian 10. But kubectx is worked as expected.

caarlos0:
Happens on macOS, iTerm (latest) and fish (latest) as well.

internal/printer/printer.go
```diff
@@ -30,16 +31,21 @@ func init() {
...
-	return err
+	return writeOutput(w, WarningColor.Sprint("warning: ")+format+"\n", args...)
 }
 
 func Success(w io.Writer, format string, args ...interface{}) error {
-	_, err := fmt.Fprintf(w, SuccessColor.Sprint("✔ ")+fmt.Sprintf(format+"\n", args...))
+	return writeOutput(w, SuccessColor.Sprint("✔ ")+format+"\n", args...)
+}
+
+func writeOutput(w io.Writer, format string, args ...interface{}) error {
```
ahmetb, line 45: do you mind if we don't do writeOutput and just do:

    \"%s\" 

at the call-site where we say "Switched [...]"? 

that way we don't have to fix and work around the color pkg. You can also a `// TODO` there saying `// TODO can't use %q there as the color sequences are escaped`

jdharmon: I don't mind. I think that is the best way forward. The only downside is replacing all the `%q%` format strings currently in code.

jdharmon:
@ahmetb I replaced all occurrences of `%q` with `"%s"`. Tested on Windows and WSL.

ahmetb:
Can you restore the deleted files?

kingdonb:
I just tripped over this, also on mac. Something to add:

```
$ kubens kube-system
✔ Active namespace is "kube-system"
kbarret8@kbarret8-mbp:~/projects/pa/personnel-actions (kubernetes-canary)$ kubens
✔ Active namespace is "\x1b[32mkube-public\x1b[0m".
```

The first invocation of "kubens" uses a parameter, prints fine.

The second invocation goes through fzf, selecting a namespace from the list. This one comes back with color codes printed as literal escape characters. I haven't been through the source to see why that might be, but it seems like it might help someone with better command of the code narrow this down faster to have better detail in the steps to reproduce.

caarlos0:
@kingdonb this patch fixes both of them, I've built my fork with it to verify - https://github.com/caarlos0/kubectx

if you want to try: `brew install caarlos0/tap/kubectx`

caarlos0:
also, @ahmetb my fork also have the needed changes to remove the bash stuff, I can open a PR with that if you want (mostly deletes and the goreleaser config).

kingdonb:
@caarlos0 It looks like your main branch does several different things, I would suggest submitting each of them separately as a different PR, since some of that might not be merged.

I'm not gonna run those binaries for example, just seems like asking to get owned, but I see your change translates `%q` to `\"%s\"` in a number of places, and I'm happy to review the source changes just as soon as I can figure out how to compile from your branch.

Matter of fact since you've done the work for me and I'm the one that thinks the commits should be reordered, let me copy your work and then go ahead and do that on my own, and I'll submit the PR. (Sec...)

kingdonb:
Confirming this issue still exists in the master branch:

```
$ kubens
✔ Active namespace is "\x1b[32mcert-manager\x1b[0m".
```
binary is built from `9527e30` `2020-11-14`

With the changes @caarlos0 mentioned, in `kingdonb/kubectx` branch: `testing`, it indeed resolves the issue:

```
$ kubens
✔ Active namespace is "dash".
```

The "dash" is green text, which you can't see here, but it looks like those control characters are being interpreted correctly. My testing was done on MacOS in iterm2.

Submitting as PR #275

jdharmon:
Sorry I didn't jump in earlier. This PR should fix the issues everyone's having.

@caarlos0 @kingdonb Can you confirm that this works correctly on Mac? I've tested on Windows and WSL.

caarlos0:
I'm using it on macOS, works great 👍

ahmetb:
Thanks!

PR #236, opened 2020-05-31
query namespace exists with GET Namespace

### LINKED ISSUE #216: kubens select and switch take too long
When using `kubens $ns` and `kubens -`, it's taking over ten seconds. This is caused by the relisting of namespaces, of which I have at least thousands. It'd be nice if this could be sped up, either by allowing me to disable the existence check or caching the namespaces.

ahmetb:
More efficient ns switches with kubens by querying only the namespace (instead
of listing all namespaces).

Fixes #216.

### DIFF
--- a/cmd/kubens/list.go
+++ b/cmd/kubens/list.go
@@ -54,15 +54,7 @@ func queryNamespaces(kc *kubeconfig.Kubeconfig) ([]string, error) {
 		return []string{"ns1", "ns2"}, nil
 	}
 
-	b, err := kc.Bytes()
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to convert in-memory kubeconfig to yaml")
-	}
-	cfg, err := clientcmd.RESTConfigFromKubeConfig(b)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


morhook:
I've tried this on my cluster (no permission to list), and got this error:

```
$ kubens my-app-namespace
error: failed to query if namespace exists (is cluster accessible?): failed to query namespace "my-app-namespace" from k8s API: namespaces "my-app-namespace" is forbidden: User "u-xxxxxxx" cannot get resource "namespaces" in API group "" in the namespace "my-app-namespace"
```

morhook:
* yes. It did work:
```
kubectl get my-app-namespace
NAME              STATUS   AGE
my-app-namespace   Active   40d
```
* yes. I do
* No. the namespace exists.

ahmetb:
OK this sounds like a bug. I am assuming you’re using the Go implementation; not bash. Can you please open an issue (and list which commit you compiled from)?

morhook:
I've just retested this. Looks like latest code doesn't produce the error. Thanks a lot for your time @ahmetb !

PR #242, opened 2020-06-13
Add version flag to show kubectx and kubens version

gauthierc:
All programs have a version option to show which version is installed. ;-)

### DIFF
--- a/cmd/kubectx/flags.go
+++ b/cmd/kubectx/flags.go
@@ -44,6 +44,9 @@ func parseArgs(argv []string) Op {
 		if v == "--unset" || v == "-u" {
 			return UnsetOp{}
 		}
+		if v == "--version" || v == "-v" {
+			return VersionOp{}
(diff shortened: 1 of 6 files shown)


gauthierc:
@googlebot I signed it!

ahmetb:
Thanks for the patch but it’s better to make a proposal first. :) We are definitely not gonna read the version from source file. Goreleaser has support for version strings.

gauthierc:
Thanks for the quick feedback. 
I didn't make a proposal for an evolution, because I found the change trivial.
I've never used Goreleaser. Indeed, it's a much better idea than putting the version number in the code.

mathew-fleisch:
I would also like a version flag. It is hard to automate updating when you cannot verify what version you are on.

ahmetb:
This is implemented (in Go version only), and will be in the next release. Stay tuned.

Issue #241, opened 2020-06-11
Proposal: Display Message When No kubectl Contexts are Available


danielhelfand:
This is a request for a very minor detail for `kubectx`. The idea would be to display a message to state explicitly that no contexts are available instead of displaying nothing as `kubectx` currently does when no contexts are available.

### Example

```
$ kubectx
No kubectl contexts available
```

ahmetb:
Yeah this is doable by printing an info message to stderr.
I'll keep in mind.

ChaturvediSulabh:
@ahmetb - May I take a gander at this?

ahmetb:
Go for it.

radhouen:
Can I start working on it ?

PR #259, opened 2020-08-18
In respond to Proposal#241- Add message when no contexts are found in kubeconfig

ChaturvediSulabh:
[See the proposal here](https://github.com/ahmetb/kubectx/issues/241) 

**What did I do?**

1. Added message for empty contexts
2. Found and removed duplicate code for "Kubeconfig not found" in lists.go, which is already handled in fzf.go. Also, it seems like fzf.go will always supersede, when there is an error as such.  
3. Tested Locally
Screenshot 2020-08-18 at 18 33 28


cmd/kubectx/list.go
```diff
@@ -19,10 +18,6 @@ func (_ ListOp) Run(stdout, stderr io.Writer) error {
 	kc := new(kubeconfig.Kubeconfig).WithLoader(kubeconfig.DefaultLoader)
 	defer kc.Close()
 	if err := kc.Parse(); err != nil {
-		if cmdutil.IsNotFoundErr(err) {
-			printer.Warning(stderr, "kubeconfig file not found")
-			return nil
-		}
```
ahmetb, line 25: why was this deleted?

ahmetb:
You need to implement integration tests as well I believe.

cmd/kubectx/list.go
ChaturvediSulabh: [Found and removed duplicate code for "Kubeconfig not found" in lists.go, which is already handled in fzf.go. Also, it seems like fzf.go will always supersede when there is an error as such.](https://github.com/ahmetb/kubectx/blob/51c61b862dcbf2293d5c037423356523e7df9e96/cmd/kubectx/fzf.go#L28)

ChaturvediSulabh:
> @googlebot I fixed it

@googlebot I fixed it

cmd/kubectx/list.go
ahmetb: but list.go#Run and fzf.go#Run are for different scenarios.
fzf.go runs if fzf is installed, list.go runs otherwise.

ChaturvediSulabh: oh! I understand now. apologies

ChaturvediSulabh:
> You need to implement integration tests as well I believe.

test case 17, delete several contexts validates this, I've updated the test case. Should I still write a separate test for this?

ahmetb:
you're right it's technically capturing this case, but it's more ideal if we copy-paste it into its own test case. technically we want to have 1:1 mapping between "behavior we want to test" and "integration test case".

cmd/kubectx/fzf.go
```diff
@@ -23,14 +23,20 @@ type InteractiveSwitchOp struct {
 func (op InteractiveSwitchOp) Run(_, stderr io.Writer) error {
 	// parse kubeconfig just to see if it can be loaded
 	kc := new(kubeconfig.Kubeconfig).WithLoader(kubeconfig.DefaultLoader)
+	defer kc.Close()
```
ahmetb, line 26: is the rest of the codebase moving defer here? I suspect if we fail doing Parse(), there's not much to close?

ChaturvediSulabh: Thanks, I realize this. Part of my learning is always to do "defer" ...

PR #261, opened 2020-08-28
Add Homebrew install for Linux

rothgar:
Is there a reason why `brew` install instructions isn't included for Linux? Because it's bash I'm able to install it just fine with brew on Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ There are several installation options:
   - Homebrew (recommended)
   - MacPorts
 - Linux
+  - Homebrew
   - Manual installation/upgrades
   - Arch Linux
   - Debian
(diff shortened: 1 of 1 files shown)


ahmetb:
Homebrew working for Linux is a relatively new thing, that's probably why. :) I am not sure about the popularity of Homebrew on Linux. If you think it is, we can add.

rothgar:
I've been using it for a while and their stats show a hundreds of thousands of package installs over the past 30 days. https://formulae.brew.sh/analytics-linux/install/30d/

No where close to the millions of macOS installs but still a fairly large number.

rothgar:
bump. I'm working on packaging some other tools and would like to include this but I'm hoping the packaging will work for macOS and linux.

ahmetb:
Nice. OK let's add.

README.md
```diff
@@ -141,6 +142,19 @@ If you use [MacPorts](https://www.macports.org) you can install like this:
 
 ### Linux
 
+#### Homebrew
```
ahmetb, line 145: Do you mind rearranging this section as:

- Debian
- Arch Linux
- Homebrew
- Manual installation

Homebrew is definitely not the #1 pkg mgr on Linux :D

ahmetb (approved):

Issue #262, opened 2020-09-10
Rename Completions Scripts for ZSH to Start with Underscore


jGleitz:
If the completion scripts started with `_`, it would be easy to install them with [antibody](https://getantibody.github.io). Simply adding 

```
ahmetb/kubectx path:completion kind:fpath
```

to `~/.zshplugins.txt` would be enough (depending on the setup, one would still need to call `compinit` in `~/.zshrc`.

This would give users a nice and easy way to install the completion script if they are already using antibody.

ahmetb:
I'm supportive, I think it might require some changes on the homebrew formula as well. (And probably alpine/debian distributions we currently have outside this repo.)

chiefy:
I'm using [zsh-snap](https://github.com/marlonrichert/zsh-snap) and having the same issue. I changed the filenames to `_kubectx` and `_kubens` and the autoloading works fine. Thanks.

PR #263, opened 2020-09-10
Make Completions Installable with Antibody

jGleitz:
Renames the zsh completions to start with an underscore and documents how to use [antibody](https://getantibody.github.io) to install them.


jGleitz:
@googlebot I signed it!

ahmetb:
I'm inclined to reject as since you can see the Readme is getting out of control wrt installation methods and I suspect Antibody isn't very popular. (Despite being a zsh user for years I heard it from you for the first time.)

jGleitz:
> Antibody isn't very popular

My feeling is the opposite: When I got started with zsh, I felt that everybody had some plugin manager and that antibody was the most popular one. But then again, I have no data to back this up.

> the Readme is getting out of control

I agree. I see two options:

 1. Use subheadings instead of only bullet points. That would make it easier to parse what is happening where.
 2. Don’t document how to install with antibody (users knowing antibody well will figure it out themselves).


My main interest is to have the scripts renamed (https://github.com/ahmetb/kubectx/issues/262) so that installing automatically via antibody is possible. I wanted to document the process for fellow antibody users, but I certainly don’t insist on that.

unixorn:
I agree with @ahmetb - anyone using a framework should know how to add a plugin. Also, if upstream changes how to install a plugin, then the information here becomes obsolete. That said, I also agree that prefixing the names with an `_` makes the completions more usable.

jGleitz:
How can we proceed? Should I remove the README changes from this PR?

>  I think it might require some changes on the homebrew formula as well.

(from #262): where would these changes need to happen?

ahmetb:
> (from #262): where would these changes need to happen?

Homebrew core repository "kubectx.rb" file.

ahmetb:
I think we can merge this. 🤞  It might still break some downstream (alpine, debian pkgs etc) but it should be manageable.

jGleitz:
@ahmetb glad that you will accept the PR :partying_face:!

To verify: Shall I remove the README changes before you merge?

ahmetb:
readme is just getting really complicated. Is there a way to implement this and keep it just as simple without adding new paragraphs?

jGleitz:
@ahmetb I have reworked the whole installation section of the `README`. From my point of view, it is now a lot easier to navigate and understand. I have linked the different sections from the overview at the beginning, I reduced the nesting, I moved the long bullet-point list into subsections and I reduced redundancies.

Does that work for you?

README.md
```diff
@@ -97,22 +97,23 @@ Active namespace is "default".
...
 ### Kubectl Plugins (macOS and Linux)
 
-You can install and use [Krew](https://github.com/kubernetes-sigs/krew/) kubectl
-plugin manager to get `kubectx` and `kubens`. **NOTE:** This will not install
-shell completion scripts, if you want those, choose another installation method
-below.
+You can install and use the [Krew](https://github.com/kubernetes-sigs/krew/) kubectl
+plugin manager to get `kubectx` and `kubens`.
+
+ **NOTE:** This will not install the shell completion scripts. If you want them, choose another installation method
```
unixorn, line 115: There's an extra space before the format markers `**Note**`

README.md
```diff
@@ -97,22 +97,23 @@ Active namespace is "default".
...
+or install the scripts [manually](#manual-installation-macos-and-linux)
```
unixorn, line 116: Please add a `.` after the link

unixorn (commented):
Found a couple very minor formatting changes.

README.md
jGleitz: Done.

README.md
jGleitz: Done.

jGleitz:
@unixorn Thank you! I fixed those.

jGleitz:
@unixcorn @ahmetb can we finally merge this? I’d still be interested in the changes.

jGleitz:
@ahmetb I fixed the merge conflict. I am looking forward to seeing this merged!

PR #269, opened 2020-10-29
fix: improve kubectx completion on fish

caarlos0:
make `kubens` and `kubectx` completion more similar to each other.


ahmetb:
Thanks much, I am not tagging a release for now.

PR #279, opened 2021-01-09
add version subcommand (using GoReleaser vars)

coolaj86:
Any of these:

```bash
kubectx version
kubectx -version 
kubectx --version
kubectx -V
```

Will produce this output:

```txt
kubectx v0.9.1-SNAPSHOT-1a50409 (1a50409) 2021-01-09T20:56:06Z
```

Also works for `kubens`.

### DIFF
--- a/cmd/kubectx/main.go
+++ b/cmd/kubectx/main.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/ahmetb/kubectx/internal/cmdutil"
 	"github.com/ahmetb/kubectx/internal/env"
@@ -29,8 +30,28 @@ type Op interface {
 	Run(stdout, stderr io.Writer) error
 }
 
+var (
+	name    = "kubectx"
+	version = "v0.0.0"
+	date    = "0001-01-01T00:00:00Z"
+	commit  = "0000000"
+)
+
+func ver() string {
+	return fmt.Sprintf("%s %s (%s) %s", name, version, commit[:7], date)
+}
+
 func main() {
 	cmdutil.PrintDeprecatedEnvWarnings(color.Error, os.Environ())
+
+	// Support [--]version and -V
+	if len(os.Args) > 1 {
+		if "version" == strings.TrimLeft(os.Args[1], "-") || "-V" == os.Args[1] {
+			fmt.Println(ver())
+			os.Exit(0)
+			return
+		}
+	}
 
 	op := parseArgs(os.Args[1:])
 	if err := op.Run(color.Output, color.Error); err != nil {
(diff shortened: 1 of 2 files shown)


ahmetb:
I think I can accept this.

* Let's reserve `-v` for something else in the future like "verbose" option.

* let's just do `-V` and nothing else, like a subcommand or --version.

* Also, let's just capture the version number and not date, or commit etc.

* Let's do it for kubens while you’re at it as well if you don’t mind.

coolaj86:
* [x] Let's reserve -v for something else in the future like "verbose" option.

I didn't add `-v`

* [ ] let's just do -V and nothing else, like a subcommand or --version.

The standard way to do it in Go is to have the `version` subcommand and Go treats `-` and `--` the same. So `-help` is the same as `--help`.

If we're doing _only_ POSIX style, it should be BOTH `-V` and `--version` (which would include `-version` unless we don't want to support normal go flag style).

If we're doing _only_ Go style it should be either `version` OR `-version` and `--version`

-  [ ] Also, let's just capture the version number and not date, or commit etc.

You're already using GoReleaser, which assumes you'll be capturing that information in some way. If you really don't want that I can take it out, but I trust provably verifiable idempotent builds (which GoReleaser is intended to provide) over "because I said so" builds.

Again, that's your call, but I suggest using at least the first 7 chars of the commit hash, and date is nice to see at-a-glance "is this recent like 6 days ago? or 6 months ago? or 6 years ago"

- [x] Let's do it for kubens while you’re at it as well if you don’t mind.

The commit includes both

Consider my rationale and let me know your final say and I'll adjust accordingly. :)

ahmetb:
> The standard way to do it in Go is to have the `version` subcommand and Go treats `-` and `--` the same. So `-help` is the same as `--help`.

This is a "program", not necessarily a "Go program". We have a bash based implementation as well, and many users are there. kubectx/kubens do not take subcommands, and I certainly don't want N ways of doing the same thing for the sake of keeping the 

> If we're doing _only_ POSIX style, it should be BOTH `-V` and `--version`

Not necessarily. I think we can live with one of them just fine, just like many programs do.

> You're already using GoReleaser, which assumes you'll be capturing that information in some way. If you really don't want that I can take it out, but I trust provably verifiable idempotent builds (which GoReleaser is intended to provide) over "because I said so" builds.

This is not how "trusted builds" work. All those variables can still be forged. Trusting software is a much more harder problem that requires verifying the integrity of even the compiler itself. You can refer to the Reflections on Trusting Trust (1984) article for more on this.

kubectx/kubens are not only distributed here on GitHub. They are still a "bash script" on many platforms. It is distributed on Homebrew, it's on Alpine repos, it's on Debian repos.

Practically none of those build/distribution systems are using GoReleaser. If you add all these variables, you’ll be making pretty much every packager’s job harder.

So I'm ok with `kubectx -V` printing a simple `vX.Y.Z` string. I don't know of any scenarios where a user would be interested in build date, git commit etc. ("it's available on GoReleaser" is not a good enough reason to use it.)

PR #288, opened 2021-02-02
Update README for installing go versions

EvyBongers:
The README states that 
> with v0.9.0 kubectx and kubens are now rewritten in Go

This PR updates the installation instructions in the README to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,15 +176,15 @@ This command will set up bash/zsh/fish completion scripts automatically.
 
 #### Manual
 
-Since `kubectx`/`kubens` are written in Bash, you should be able to install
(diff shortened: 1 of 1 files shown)


ahmetb:
Thanks for the patch but we won't recommend go-get as a download mechanism. 

The old instructions still apply (download from releases, then chmod+x). Just need to adjust the language around Bash. 

Also Bash is still the main version distributed via most package managers. It's the most stable implementation. Go version still has some shortcomings.

When Go becomes the primarily recommended version, I'll update the README.

EvyBongers:
Thank you for explaining.

PR #293, opened 2021-03-01
store the new context name

scones:
.. to use with other scripts or the PS1

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -108,6 +108,7 @@ save_context() {
 }
 
 switch_context() {
+  echo "${1}" > $HOME/.kube-context
   $KUBECTL config use-context "${1}"
 }
 


ahmetb:
You can use kubectl to get it.

scones:
yes, one can. i also can send pidgeons to aws with and encrypted stick to a dedicated staff member to decrypt, extract, re-encrypt and send back the information.

or i can just cache it locally, as it only changes on request.

this is a optimization feature, not a functional one.

ahmetb:
There already is a file on disk that has the context file name. So what you’re trying to do is not only redundant, it creates multiple sources of truth for the same piece of information. This info doesn’t only change via kubectx. I recommend you examine the source code of the script to understand how it works.

Issue #254, opened 2020-07-10
[Go] shouldn't use XDG_CACHE_HOME for reading kubeconfigs


ahmetb:
bash based implementations used something like `KUBECTX="${XDG_CACHE_HOME:-$HOME/.kube}/kubectx"` to determine where the "state files" are stored.

XDG_CACHE_HOME is good for that kinda stuff.

However with the Go implementations, that path is also used for determining where the $KUBECONFIG file is at by default –which is not correct. client-go default config loading rules don't read XDG_ env vars.

josegonzalez:
I think this can be closed.

aslafy-z:
I don't think this issue should be closed: The Go implementation stills push cache files to the `$HOME/.kube` directory: https://github.com/ahmetb/kubectx/blob/master/cmd/kubectx/state.go#L27-L33.

PR #299, opened 2021-05-27
stop using XDG_CACHE_HOME as home directory

pjjw:
XDG_CACHE_HOME is not a substitute for $HOME, see [1].

[1]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

### DIFF
--- a/internal/cmdutil/util.go
+++ b/internal/cmdutil/util.go
@@ -21,9 +21,6 @@ import (
 )
 
 func HomeDir() string {
-	if v := os.Getenv("XDG_CACHE_HOME"); v != "" {
-		return v
-	}
 	home := os.Getenv("HOME")
 	if home == "" {
 		home = os.Getenv("USERPROFILE") // windows
--- a/internal/cmdutil/util_test.go
+++ b/internal/cmdutil/util_test.go
@@ -28,12 +28,12 @@ func Test_homeDir(t *testing.T) {
 		want string
 	}{
 		{
-			name: "XDG_CACHE_HOME precedence",
+			name: "don't use XDG_CACHE_HOME as homedir",
 			envs: []env{
 				{"XDG_CACHE_HOME", "xdg"},
 				{"HOME", "home"},
 			},
-			want: "xdg",
+			want: "home",
 		},
 		{
 			name: "HOME over USERPROFILE",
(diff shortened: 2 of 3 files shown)


ahmetb:
addresses #254, right?

pjjw:
yep! that's the idea.

pjjw:
i can't replicate the CI failure locally, and CI also appears to be suppressing the error messages. Here's local:
```
[16:36:49][pjjw@loon:~/src/kubectx(pjjw/fix-cmdutil-homedir)]$ go build cmd/
[16:36:56][pjjw@loon:~/src/kubectx(pjjw/fix-cmdutil-homedir)]$ unset XDG_CACHE_HOME
[16:36:57][pjjw@loon:~/src/kubectx(pjjw/fix-cmdutil-homedir)]$ COMMAND=$(PWD)/kubens ./node_modules/.bin/bats test/kubens.bats 
 ✓ --help should not fail
 ✓ -h should not fail
 ✓ list namespaces when no kubeconfig exists
 ✓ list namespaces
 ✓ switch to existing namespace
 ✓ switch to non-existing namespace
 ✓ switch between namespaces
 ✓ switch to previous namespace when none exists
 ✓ switch to namespace when current context is empty
 ✓ -c/--current works when no namespace is set on context
 ✓ -c/--current prints the namespace after it is set
 ✓ -c/--current fails when current context is not set

12 tests, 0 failures

[16:37:03][pjjw@loon:~/src/kubectx(pjjw/fix-cmdutil-homedir)]$ XDG_CACHE_HOME=$HOME/.cache COMMAND=$(PWD)/kubens ./node_modules/.bin/bats test/kubens.bats 
 ✓ --help should not fail
 ✓ -h should not fail
 ✓ list namespaces when no kubeconfig exists
 ✓ list namespaces
 ✓ switch to existing namespace
 ✓ switch to non-existing namespace
 ✓ switch between namespaces
 ✓ switch to previous namespace when none exists
 ✓ switch to namespace when current context is empty
 ✓ -c/--current works when no namespace is set on context
 ✓ -c/--current prints the namespace after it is set
 ✓ -c/--current fails when current context is not set

12 tests, 0 failures
```

pjjw:
okay, got it, i think. tests were only working because $XDG_CACHE_HOME counted as $HOME. PTAL.

ahmetb:
FWIW I think we had this XDG_ logic early in the bash implementation (a long time ago now). I can't recall why we added it but it's definitely non-standard. I think we just ported it over to Go implementation (and therefore kept the tests the same).

pjjw:
anything else needed for this to land?

ahmetb:
Thank you. 🙏

hcc-clarity:
can we have a release with this fix merged?

ahmetb:
Fixed in v0.9.4.

PR #300, opened 2021-06-03
386, s390x and risv64 CPU archs for releases

qdm12:
- 386 can be actually useful, especially on (cheap) NAS with 32 bit non-ARM processors
- s390x is for the fun of it AFAIK
- riscv64 is really not that useful but could be in the future, and it's supported by goreleaser since January 2021 😉

Thanks!

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -33,8 +33,6 @@ builds:
     - arm
     - arm64
     - ppc64le
-    - s390x
-    - riscv64
   goarm: [6, 7]
 - id: kubens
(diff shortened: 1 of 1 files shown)


ahmetb:
We only accept the new platform submissions if there are users requesting to use them, not based on hypothetical needs. Please close this issue if you actually do not need them.

qdm12:
As mentioned above, 386 is definitely useful to me, s390x and riscv64 are not. I will modify the PR to only add 386.

PR #308, opened 2021-07-20
removing duplicate homebrew section from README.md

tminussi:


ahmetb:
this is somewhat intentional since both mac and linux support homebrew.

Issue #305, opened 2021-07-13
skim support


aschleifer:
Would it be possible to introduce [skim](https://github.com/lotabout/skim) support to the tools similar to the already existing fzf support?

ahmetb:
That's worth considering. We can expose a FINDER or PICKER environment variable for that and set its implicit default to `fzf --ansi --no-preview`. It seems `sk` offers identical options as well.

When I tried it hands-on for a minute, I observed no differences between fzf and sk. That's why I'm a bit hesitant, but both seem popular enough that maybe we can offer a knob. That said, this never came up until today.

Ant0wan:
That would be a nice feature ! :+1:

PR #312, opened 2021-08-08
Adding skim support

frzam:
[See the Feature request here](https://github.com/ahmetb/kubectx/issues/305)

Added skim support in kubens and kubectx go version.

Added environment variable **PICKER** which will contain value **sk** or **fzf**, if PICKER is not set or anything else is set other than sk and fzf, then it will by default pick fzf.

Created a func InteractiveSearch in interactive.go which will be called for fuzzy search and renamed fzf.go to fuzzy.go since it is now more generic.

### DIFF
--- a/internal/env/constants.go
+++ b/internal/env/constants.go
@@ -29,4 +29,12 @@ const (
 
 	// EnvDebug describes the internal environment variable for more verbose logging.
 	EnvDebug = `DEBUG`
+
+	// EnvPicker describes the environment variable for fuzzy support, It can value
+	// fzf or sk. If this is not set then fzf is taken as default picker.
+	EnvPicker = `PICKER`
+
+	// EnvSKIgnore describes the environment variable to disable interactive context
+	// selection when skim is installed
+	EnvSKIgnore = `KUBECTX_IGNORE_SK`
 )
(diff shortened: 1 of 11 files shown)


frzam:
@googlebot I signed it!

ahmetb:
Hey, sorry I've been slow reviewing things!
For compatibility purposes, do you mind if we check `fzf` first, then `sk`? I am suspecting if people have another binary somehow called sk in their PATH, their workflow might be broken. (And this can work because I assume sk users don't have fzf installed?)

internal/env/constants.go
```diff
@@ -29,4 +29,12 @@ const (
 
 	// EnvDebug describes the internal environment variable for more verbose logging.
 	EnvDebug = `DEBUG`
+
+	// EnvPicker describes the environment variable for fuzzy support, It can value
+	// fzf or sk. If this is not set then fzf is taken as default picker.
```
ahmetb, line 34: > If this is not set then fzf is taken as default picker.

My expectation of this feature was that people can set this to anything and we can feature-proof the tool. e.g. imagine a tool named `abc` is developed in the future, I should be able to say `PICKER=abc --opt1 -opt2`.

I think this is how more common env vars like EDITOR or PAGER in posix systems work. The caller has no information about the program it's calling other than it knows how to pass it the information (either via xargs or via stdin).

ahmetb: Also this way, we can retain fzf logic and have an override via `PICKER`. The source code does not need to know about `sk` at all.

sambonbonne:
For those who are waiting for this PR to be merged, you cna create an alias with `kubectx $(kubectx | sk)`, it's not a pefect solution but it works.

PR #326, opened 2021-10-18
Removed duplicate information in README

biniama:
Removed duplicate information in README about installation instruction via Homebrew


ahmetb:
We reorganized this file a bit already. But it was duplicate because Homebrew is valid both for Linux as well as macOS.

PR #331, opened 2021-11-15
Added hability to switch to a context using a partial name

jordiprats:
This PR is to be able to switch to the context if only one matches. This is intended for EKS clusters where the default context name is the ARN (arn:aws:eks:REGION:ACCOUNTNUMBER:cluster/CLUSTERNAME) which is quite long but includes the cluster name

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -108,7 +108,11 @@ save_context() {
 }
 
 switch_context() {
-  $KUBECTL config use-context "${1}"
+  if [ "$(get_contexts | grep -c "${1}")" -eq 1 ]; then
+    $KUBECTL config use-context "$(get_contexts | grep "${1}")"
+  else
+    $KUBECTL config use-context "${1}"
+  fi
 }
 
 choose_context_interactive() {


jordiprats:
@googlebot I signed it!

ahmetb:
This did come up in the past and I think (1) it's kinda dangerous as that might not always be the intent and might switch into an unintended context, especially in a non-interactive setting (2) fzf has a builtin flag/feature that does this (I believe "--select-1") (3) we are accepting new changes to the Go implementation at this point (4) there's #324 adding this (with fzf support). That's why I'm inclined to not pursue this.

jordiprats:
Would it be this accepted if I add this as a flag?

For example:

kubectl ctx -s example

ahmetb:
I think the fzf support already is going to offer this in the PR I linked so I don't think this should be a separate option like -s. 

Something like this makes sense only in the interactive context (i.e. not from a script) and in that scenario, we already have a fuzzy search integration with fzf for a long time now and we'll add tools like sk as well (there's an open issue on that too).

So I am incl￼ined to not accept this as a new option or to the bash script. The bash based script is rather small and static , feel free to maintain your fork.

jordiprats:
Sure, thank you for your feedback. I don't really like it to be interactive, that's why I have proposed this

PR #334, opened 2021-11-24
Display fzf when provided with an invalid context

pschmitt:


anton-ryzhov:
I'd love to have this feature, it's very convenient to use partial identifiers

ahmetb:
not interested in implementing a fallback mode. it's inherently unintuitive for a tool to do something different given the same input. if fzf were to auto-accept for example, it might switch to a context that was not intended.

given there's already a fzf integration that can do prefix search, I'm inclined to decline it (+bash implementation is frozen at this point).

PR #344, opened 2022-01-28
Support deleting multiple contexts separated by spaces

suzaku:
In the current implementation, `kubectx -d ctx1 ctx2` doesn't work because it's interpreted as `'ctx1 ctx2'`.


ahmetb:
This is a wontfix for the time being. You can work around it with `| xargs -n 1 kubectx -d` if you want.

Issue #348, opened 2022-03-07
Option to unset the namespace for the current context


mamiu:
## Feature Request

The `kubectx` command has the flag `-u` / `--unset` to unset the current context.
This is currently not possible with the `kubens` command.

Please add a flag like `-u` (for unset) or `-x` (for delete / reset) to the `kubens` command to unset the namespace configured with `kubens`.

ATM this is only possible with the following kubectl command:

```bash
kubectl config unset contexts.<NAME-OF-THE-CURRENT-CONTEXT>.namespace
```

## Reproduce

```bash
kubectl config get-contexts

# Output:
CURRENT   NAME      CLUSTER   AUTHINFO        NAMESPACE
*         example   example   admin@example
```

```bash
kubens demo-namespace

CURRENT   NAME      CLUSTER   AUTHINFO        NAMESPACE
*         example   example   admin@example   demo-namespace
```

Now it's impossible to reset the set namespace. Setting it to "default" also won't unset it but just set it to the "default" namespace.

The only option to reset the namespace is to unset it with the following kubectl command:

```bash
kubectl config unset contexts.example.namespace
```

Akaame:
I have created a crude PR for implementing this behavior. Howeverm now the feature is here, I am wondering what is the concrete benefit of this for you over setting it to "default"?

mamiu:
Wow, thanks a lot @Akaame!

### The problem:
Applications like [k9s](https://github.com/derailed/k9s) will use the namespace set for a particular context by default every time you start them but won't adjust the namespace for that context even when you switch between namespaces (probably because it's a change to the kube config file and I guess that's something tools like k9s and Lens don't wanna touch). K9s always used to go to the last used namespace on open (which is stored in its config file), but now always goes to the namespace specified in the kube config file (which was set by `kubens`). Since I'm using k9s most of the time and the `kubectl` CLI (with `kubectx` and `kubens`) only from time to time, it always requires an extra step of switching to the desired namespace in k9s.

### The benefit:
Having the option to just unset the namespace for a context with `kubens -u` is an easy and memorable step at the end of using the `kubectl` toolchain. Third party tools then are independent from the namespace setting in the kube config file.

Akaame:
Okay. I also had the same problem when I was switching over from k9s to kubectx+kubens+fzf.

[...]

PR #349, opened 2022-03-15
348: initial impl for unsetting kubens

Akaame:
Initial draft for #348


ahmetb:
I think the idea sounds good. Let's also add the necessary tests (integration/unit) and update the README.
This will create some discrepancy between kubens bash version and the Go one, but that's ok.

PR #352, opened 2022-03-28
Add support for Snap

### LINKED ISSUE #351: Snap support
I'm developing a Snap for kubectx+kubens and I have a functional version. I will post a PR shortly, but some points still remains:
- It needs an interface connection, which isn't automatic. The idea is to open a request to enable the auto-connection for this interface
- It requires an alias. The Snap is called "kubectx", but it has two commands: `kubectx` and `kubens`. The first one works fine, just type `kubectx` and that is it, but the second one, however, requires more effort. You need to type `kubectx.kubens`. An alias can be requested to just type `kubens`, but again, a request has to be done
- A third question is we could try to use to tracks, one for go version and another one for bash version, but more research needs to be done about that
- I couldn't make the auto-completion work on my machine, but I use Zsh, not Bash, so help will be appreciated.

mhalano:
Add support for Snap to Go version

Closes #351

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,41 @@
+name: kubectx
+base: core20
+version: '0.9.4'
+summary: 'kubectx + kubens: Power tools for kubectl'
+description: |
+  kubectx is a tool to switch between contexts (clusters) on kubectl faster.
+  kubens is a tool to switch between Kubernetes namespaces (and configure them for kubectl) easily.
+grade: stable
+confinement: strict
+
+plugs:
+  dot-kube-config:
+    interface: personal-files
+    write: [$HOME/.kube/config]
+
+apps:
+  kubectx:
+    command: bin/kubectx
+    completer: completion/kubectx.bash
+    plugs:
+      - dot-kube-config
+    environment:
+      HOME: $SNAP_REAL_HOME
+  kubens:
+    command: bin/kubens
+    completer: completion/kubens.bash
+    plugs:
+    - dot-kube-config
+    environment:
+      HOME: $SNAP_REAL_HOME
+
+parts:
+  kubectx-kubens:
+    source: .
+    plugin: go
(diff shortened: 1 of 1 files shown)


mhalano:
1. Snap is a package format with a strong confinement feature, very similar in concept to Flatpak. Snap is a Canonical's initiative. Canonical is the company behind Ubuntu. Here is a link with additional information: https://en.wikipedia.org/wiki/Snap_(software)
2. I think I can find a way to get the version in another place, I just need to know where is this place inside the source code.
3. Yes, there is. But the idea is to integrate. I could create a personal repository just with snapcraft.yaml, it would work too. But may be I could convert to a .goreleaser.yaml file. Maybe it could be better. What do you think?

mhalano:
@ahmetb The Snap is working 100% I did some requests, and they were granted. My question is if you want I use goreleaser instead of pure snapcraft.yaml file.

ahmetb:
We already have Goreleaser set up. So I prefer that over maintaining version number hardcoded. How does this work? Does it upload binaries etc?

mhalano:
You generate a Snap using the command 'snapcraft' and to the upload to the Snap Store. Yes, 'snapcraft' command build your application in a clean environment (usually a VM with another tool called multipass) and generates this file .snap which you (in this case I) have to upload to the official store.

mhalano:
@ahmetb Do I have your bless to integrate the snapcraft.yaml with the .goreleaser.yaml? If yes, I will work on that this weekend and send a new MR ASAP

ahmetb:
Yeah go for it. Curious how it'll look like.

mhalano:
@ahmetb I created a new commit with the Snap implementation using Goreleaser. I think it's even better because now it's possible to publish mutli-arch Snaps with a major release or a regular snapshot.

mhalano:
Closing this PR to keep the conversation in PR #353

PR #353, opened 2022-04-09
Add support to publish Snaps using goreleaser

### LINKED ISSUE #351: Snap support

mhalano:
Add support to publish Snaps using goreleaser

Closes #351

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       uses: goreleaser/goreleaser-action@v2
       with:
         version: latest
-        args: release --snapshot --skip-publish --rm-dist
+        args: release --snapshot --skip publish,snapcraft --clean
     - name: Setup BATS framework
       run: sudo npm install -g bats
     - name: kubectx (Go) integration tests
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,12 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: '1.20'
+    - name: Install Snapcraft
+      uses: samuelmeuli/action-snapcraft@v1
+    - name: Setup Snapcraft
+      run: |
+        mkdir -p $HOME/.cache/snapcraft/download
+        mkdir -p $HOME/.cache/snapcraft/stage-packages
     - name: GoReleaser
       uses: goreleaser/goreleaser-action@v2
       with:
@@ -42,3 +48,7 @@ jobs:
       uses: rajatjindal/krew-release-bot@v0.0.38
       with:
         krew_template_file: .krew/ns.yaml
+    - name: Publish Snaps to the Snap Store (stable channel)
+      run: for snap in $(ls dist/*.snap); do snapcraft upload --release=stable $snap; done
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,7 @@ jobs:
       uses: samuelmeuli/action-snapcraft@v1
     - name: Setup Snapcraft
       run: |
+        # https://github.com/goreleaser/goreleaser/issues/1715
         mkdir -p $HOME/.cache/snapcraft/download
         mkdir -p $HOME/.cache/snapcraft/stage-packages
     - name: GoReleaser
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,4 +51,4 @@ jobs:
     - name: Publish Snaps to the Snap Store (stable channel)
       run: for snap in $(ls dist/*.snap); do snapcraft upload --release=stable $snap; done
       env:
-        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -98,3 +98,20 @@ release:
   extra_files:
     - glob: ./kubens
     - glob: ./kubectx
+snapcrafts:
+  - id: kubectx
+    name: kubectx
+    summary: 'kubectx + kubens: Power tools for kubectl'
+    description: |
+      kubectx is a tool to switch between contexts (clusters) on kubectl faster.
+      kubens is a tool to switch between Kubernetes namespaces (and configure them for kubectl) easily.
+    grade: stable
+    confinement: classic
+    base: core20
+    apps:
+      kubectx:
+        command: kubectx
+        completer: completion/kubectx.bash
+      kubens:
+        command: kubens
+        completer: completion/kubens.bash


.goreleaser.yml
```diff
@@ -92,3 +92,32 @@ release:
...
+    description: |
+      kubectx is a tool to switch between contexts (clusters) on kubectl faster.
+      kubens is a tool to switch between Kubernetes namespaces (and configure them for kubectl) easily.
+    grade: stable
+    confinement: strict
+    base: core20
+    plugs:
+      dot-kube-config:
+        interface: personal-files
+        write: [$HOME/.kube/config]
```
ahmetb, line 108: Some people specify custom paths with >1 files. How would this work here?

.goreleaser.yml
```diff
@@ -92,3 +92,32 @@ release:
...
+    apps:
+      kubectx:
+        command: kubectx
+        completer: completion/kubectx.bash
+        plugs:
+        - dot-kube-config
+        environment:
```
ahmetb, line 115: Probably need $KUBECONFIG here as well.

ahmetb (commented):
Thank you. 🙏🏼

mhalano:
I need to do some review in the code. I think I will need a special authorization to use the classic confinement instead of strict one. I already submit the request and if and when I get it, I will do changes in the code. Please stay tuned.

ahmetb:
Any updates?

mhalano:
> Any updates?

@ahmetb No updates yet. I need what we call a 'classic' confinement to snap, but it has been a little hard to get an answer about my request.

mhalano:
@ahmetb Just to keep you on the loop: I'm still working on it. They answer me back some days ago, and now I'm trying to a new line of action to be approved. Permissions with Snap can be hard, but I think I will make it. Stay tuned.

mhalano:
@ahmetb The process to get (or not) the final approval for classic confinement is almost done. I already submitted a version with the classic confinement to the store, and I will make the final code and submit the PR to your repo ASAP.

ahmetb:
Thanks. How do we do that? Is that something Goreleaser does or we need to authenticate and upload a different way?

mhalano:
We can use GitHub Actions. Do you already use it to publish the assets?

ahmetb:
@mhalano yeah, we upload to GitHub Releases using GitHub Actions. What's involved here? Do we need to call an external API to upload assets?

mhalano:
We need the command `snapcraft` in the PATH of builder (container or VM) so it can build the Snapcraft. I will look about more details how to add a step to publish the .snap file generated by the Goreleaser.

mhalano:
I sent to your email an invitation to manage with me the Snap in the Snap Store.

[Here](https://snapcraft.io/kubectx/builds) there is a build system in place integrated with GitHub, but the problem it demands a `snapcraft.yaml`, so we need to take the long way and do the commands manually to use `snapcraft` to login and push the assets.

ahmetb:
> so we need to take the long way and do the commands manually to use `snapcraft` to login and push the assets.

At that point I might question do we actually need this distribution method. So far I am spending zero effort on releases in the upstream repo and anyone packaging this software in apt/yum like repos are regularly picking up the releases from here.

> very interesting page about a GitHub Action that could solve the problem

Yeah, if we can't automate it with a GitHub Action then we probably should leave it out of this upstream repo. I'm happy to add a token or whatever to GitHub repo secrets, but you should probably first test this on your fork and send the changes here.

mhalano:
@ahmetb I added a commit that adds the Snapcraft and now it's building correctly. I just got a problem with bats execution. Can you help me with that?

mhalano:
BTW: How would you implement secrets, so you can add the Snap Store token to upload the Snap automatically? I read you must use environments to implement secrets, but I don't know if you prefer to use a simple environment variable instead of a secret (which needs an environment configured).

mhalano:
I think it's done. You will need to create an Ubuntu SSO account and to install `snapcraft` tool to get the token with this command:
```
snapcraft export-login --snaps kubectx --channels "edge stable" --acls package_upload -
```
When you execute this command, it will open a webpage to login with your Ubuntu SSO account. That should be a no brainer.

ahmetb:
@mhalano I've started to think this alternative approach:

What if you had a separate ~empty repository where you have a GitHub Action that runs once per day to see if there's a new tag on `ahmetb/kubectx` (by keeping track of last seen version in git), and when there's a new release, you'd wget the published assets and upload a snap release from your repo? That way:

1. we keep build times low in this repo
2. we don't have to maintain per-distribution-method details in the upstream repo (I don't think we have any today, despite kubectx is packaged for many package managers)
3. you can better manage what goes to Snap (description, versions etc) by taking me out of the loop as you'd be the packager.

Executing on this possibility probably depends on what goreleaser spits out here and whether you can repeat the same functionality on a separate repo.

.github/workflows/ci.yml
```diff
@@ -57,3 +63,9 @@ jobs:
       run: COMMAND=./dist/kubectx_linux_amd64/kubectx bats test/kubectx.bats
     - name: kubens (Go) integration tests
       run: COMMAND=./dist/kubens_linux_amd64/kubens bats test/kubens.bats
+    - name: Publish Snaps to the Snap Store (edge channel)
+      run: for snap in $(ls dist/*.snap); do snapcraft upload --release=edge $snap; done
+      env:
+        SNAPCRAFT_STORE_AUTH: candid
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_TOKEN }}
```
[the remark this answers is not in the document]
mhalano: Ok, I will remove it. The idea was to publish an edge version of the code, but if you prefer to keep just with stable versions I can do that.

.github/workflows/release.yml
```diff
@@ -27,6 +27,12 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1.17
+    - name: Setup Snapcraft
```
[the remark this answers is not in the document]
mhalano: I don't think it will be possible. This part makes available the snapcraft command. And it need to be executed before goreleaser. The other part publishes the snaps that already had builded.

.github/workflows/ci.yml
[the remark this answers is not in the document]
mhalano: Removed

mhalano:
@ahmetb I really like the way we are going to. I think the use of Goreleaser is a nice approach and in the future we could have packaging for other formats. I think this is important because sometimes if you depend on distro-packaging, like in kubectx case, to get the software installed nicely, using a package management solution, you probably need to sit and wait, dying in the process. I created the Snap version because I use kubectx/kubens and like them a lot, but Debian still didn't package and Ubuntu neither. Snap is nicer because is cross-distro and easier to create than Flatpak

mhalano:
> > I just got a problem with bats execution in my local machine because absence of assets.
> 
> you should be able to run `bats` locally or in your fork. maybe things broke since the last time we ran it.

The problem seems to be the directory name. The directory present inside `dist/` is called `kubectx_linux_amd64_v1` with `_v1` in the end.

ahmetb:
Ah it seems like goreleaser itself has changed. No idea where that _v1 comes from. master branch's probably broken too.

.github/workflows/ci.yml
```diff
@@ -26,6 +26,12 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: 1.17
+    - name: Setup Snapcraft
+      uses: samuelmeuli/action-snapcraft@v1
+    - id: create-snapcraft-cache-dirs
+      run: |
+        mkdir -p $HOME/.cache/snapcraft/download
+        mkdir -p $HOME/.cache/snapcraft/stage-packages
```
ahmetb, line 34: why install if not used?

ahmetb: nvm I think goreleaser uses it.

ahmetb:
opened https://github.com/goreleaser/goreleaser/issues/3192 to track the binary name change issue, but feel free to send a PR to master pinning goreleaser version to 1.7.0 for the time being.

mhalano:
@ahmetb I added a commit to change Goreleaser version from latest to 1.7.0 and at least the CI pipeline worked flawlessly (I didn't try the release pipeline). I read the answer for your Goreleaser issue and it seems a quite tricky problem to solve. Not impossible, just tricky.

mhalano:
@ahmetb please review my final PR. I also released on Snap Store the version 0.9.5 of kubectx.

.github/workflows/ci.yml
```diff
@@ -57,3 +65,7 @@ jobs:
       run: COMMAND=./dist/kubectx_linux_amd64_v1/kubectx bats test/kubectx.bats
     - name: kubens (Go) integration tests
       run: COMMAND=./dist/kubens_linux_amd64_v1/kubens bats test/kubens.bats
+    - name: Publish Snaps to the Snap Store (edge channel)
+      run: for snap in $(ls ./dist/*.snap); do snapcraft upload --release=edge $snap; done
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
```
ahmetb, line 71: I think we should reuse the same name that makes it clear the secret is for snapcraft. 

I assume you'd be delivering the secret to me somehow? Would it be possible to list me as co-maintainer too? Do I need to open an account?

.github/workflows/ci.yml
```diff
@@ -57,3 +65,7 @@ jobs:
       run: COMMAND=./dist/kubectx_linux_amd64_v1/kubectx bats test/kubectx.bats
     - name: kubens (Go) integration tests
       run: COMMAND=./dist/kubens_linux_amd64_v1/kubens bats test/kubens.bats
+    - name: Publish Snaps to the Snap Store (edge channel)
```
ahmetb, line 68: Why is this step in "ci"? Shouldn't it be release-only?

.github/workflows/ci.yml
mhalano: What name do you want? I used the same on the documentation I found.
Related to the token value, you will need to create an account here: https://login.ubuntu.com. I will give permissions to your user to manage the application, or even transfer the application to you if you want so. After that, you need to get a token using the command like `snapcraft export-login --snaps=kubectx  --acls package_access,package_push,package_update,package_release credentials.txt` and use the content of the file as the value in the variable.

.github/workflows/ci.yml
mhalano: It's to publish a SNAPSHOT release on edge channel with each new pipeline run. It's very useful to debug things, but I can remove and just keep the main run on release.yaml. Your call.

mhalano: image

ahmetb: But this will run even on unmerged PR tests. I don't think that's safe or trusted.

mhalano: Ok then. I will remove. Give 1 min.

mhalano: Done. Now it will just run with release.yaml. Any other considerations?

.github/workflows/ci.yml
mhalano: Btw, I checked your email is already added.
image

.github/workflows/release.yml
```diff
@@ -42,3 +48,7 @@ jobs:
       uses: rajatjindal/krew-release-bot@v0.0.38
       with:
         krew_template_file: .krew/ns.yaml
+    - name: Publish Snaps to the Snap Store (stable channel)
+      run: for snap in $(ls dist/*.snap); do snapcraft upload --release=stable $snap; done
+      env:
+        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.STORE_LOGIN }}
```
ahmetb, line 54: I just realized I added a `$SNAPCRAFT_TOKEN` to the repo last year. Can we just use that?

.github/workflows/release.yml
```diff
@@ -27,6 +27,12 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: '1.20'
+    - name: Install Snapcraft
+      uses: samuelmeuli/action-snapcraft@v1
+    - name: Setup Snapcraft
+      run: |
+        mkdir -p $HOME/.cache/snapcraft/download
+        mkdir -p $HOME/.cache/snapcraft/stage-packages
```
ahmetb, line 35: This step feels odd. Why doesn't the action above (or the `snap` command create these)? It doesn't seem you're using this cache between runs either?  (we don't need, just asking.)

.github/workflows/release.yml
mhalano: Done

.github/workflows/release.yml
mhalano: These couple lines are necessary because of [this](https://github.com/goreleaser/goreleaser/issues/1715) problem. If it doesn't exist when snapcraft command is executed things go wild. Kidding. It just crash and burns. :)

mhalano: I had one more link about this as a comment in the file, but I think I lost that commit. :/

ahmetb: Yeah, let's add a comment as to why we're adding this step. I definitely won't remember this in a few months :)

mhalano: Done

mhalano: @ahmetb Do you have any other considerations?

ahmetb: I clicked on "Run action" and CI seems to be failing. I think we need a way to configure goreleaser to skip snapcraft step while just building/packaging (and skip publishing).

mhalano: I saw it. I already have an idea

mhalano: @ahmetb Done. Please check again.

ahmetb:
Thanks for the patience here. It’ll go out with the new release, hopefully not too far from now.

PR #356, opened 2022-04-28
add search feature

jessecorson:
Add search feature
- example usage `kubectl ctx -s foo bar` will return contexts with both foo and bar in them which can then be selected

### DIFF
--- a/kubectx
+++ b/kubectx
@@ -21,7 +21,11 @@
 set -eou pipefail
 IFS=$'\n\t'
 
-SELF_CMD="$0"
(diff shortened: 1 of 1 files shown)


jessecorson:
@ahmetb it didn't take me very long to write, but a search feature would be nice, if you are listing your contexts and have hundreds of them, you might not want to up arrow through all of them or copy and paste.

jessecorson:
@ahmetb it didn't take long to write, but now I know that fzf has search so that pretty much takes care of it! thanks

PR #360, opened 2022-08-02
Bump golang.org/x/sys to fix compilation on M1 macs

SuperSandro2000:


ahmetb:
Was the compilation broken? I believe tests cover building binaries at each commit for all platforms.

SuperSandro2000:
Yes, on go 1.18 for m1 macs.

see https://github.com/google/go-jsonnet/issues/596 for the same issue just on another repository

Issue #157, opened 2019-06-15
Completions don't work


jwenz723:
I have installed kubectx and kubens using the brew installation `brew install kubectx`. The README states:

> This command will set up bash/zsh/fish completion scripts automatically.

The installation succeeded, and I am able to use both kubectx and kubens, however completions are not working (assuming I'm doing it right).  I type `kubectx esc` (where esc is the first 3 letters of my context that is returned by running `kubectx`) then hit `<tab><tab>` and nothing happens.

If I type `kubectx ` then hit `<tab><tab>` my shell just starts running through all the directories in my current path (pwd).

I'm on MacOS Mojave (10.14.5)
My shell is oh-my-zsh with robbyrussell theme

I do see that the completion scripts are installed by brew in a location that is on `fpath`:

```
➜  ~  echo $fpath
/Users/jeff.wenzbauer/.oh-my-zsh/plugins/kube-ps1 /Users/jeff.wenzbauer/.oh-my-zsh/plugins/kubectl /Users/jeff.wenzbauer/.oh-my-zsh/plugins/history /Users/jeff.wenzbauer/.oh-my-zsh/plugins/git /Users/jeff.wenzbauer/.oh-my-zsh/functions /Users/jeff.wenzbauer/.oh-my-zsh/completions /usr/local/share/zsh/site-functions /usr/share/zsh/site-functions /usr/share/zsh/5.3/functions
➜  ~  ls /usr/local/share/zsh/site-functions
_brew               _brew_cask          _brew_services      _git                _helm               _kubectl            _kubectx            _kubens             git-completion.bash linkerd
```

ahmetb:
If you search "completions" in the issue tracker, there are many issues that came up in the. Maybe you're not using `zsh-completions` plugin in oh-my-zsh. Maybe you need to do a `autoload -U compinit && compinit`. Hard to tell...

I'm inclined to close this, as it probably works for hundreds of other users out there, and likely a setup issue with you.

trondhindenes:
`autoload -U compinit && compinit` was what was missing for me.

jwenz723:
Yep, that was the problem. I had to install [zsh-completions](https://github.com/zsh-users/zsh-completions#using-zsh-frameworks) and then add into ~/.zshrc:

```
plugins=(… zsh-completions)
autoload -U compinit && compinit
```

zroupas:
[...]

PR #368, opened 2022-10-22
Add prerequisite for completion to work after brew install

vincentvictoria:
I bumped into [Completions don't work](https://github.com/ahmetb/kubectx/issues/157#issuecomment-1287588271) issue, and found a root cause: https://github.com/ahmetb/kubectx/issues/157#issuecomment-1287588271.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ If you use [Homebrew](https://brew.sh/) you can install like this:
 brew install kubectx
 ```
 
-This command will set up bash/zsh/fish completion scripts automatically. However, you must [setup completion for brew](https://docs.brew.sh/Shell-Completion) for it to work.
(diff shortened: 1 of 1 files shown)


ahmetb:
Is this not true for anything else installed via Brew? It feels a bit awkward that every project has to document it.

README.md
```diff
@@ -109,7 +109,7 @@ If you use [Homebrew](https://brew.sh/) you can install like this:
 brew install kubectx
 ```
 
-This command will set up bash/zsh/fish completion scripts automatically.
+This command will set up bash/zsh/fish completion scripts automatically. However, you must [setup completion for brew](https://docs.brew.sh/Shell-Completion) for it to work.
```
ahmetb, line 112: Let's change this too.

> Make sure you [configure your shell] to pick up completions for Homebrew formulas.

vincentvictoria:
> Is this not true for anything else installed via Brew? It feels a bit awkward that every project has to document it.

That's true, but it will help people who haven't done so. It's probably better for Brew to put a link on their front page along with the one-liner install script.

Issue #285, opened 2021-01-27
_arguments:comparguments:325: can only be called from completion function


martinmaehlmann:
### The zsh completion scripts make zsh throw an error.

`_arguments:comparguments:325: can only be called from completion function`

This is due to the fact, that both zsh completion scripts call `_arguments` outside a function.

### Fix

Put the completion scripts into a function, as shown below:

**kubectx.zsh**
```
 #compdef kubectx kctx=kubectx

__kubectx() {
  local KUBECTX="${HOME}/.kube/kubectx"
  PREV=""

  local all_contexts="$(kubectl config get-contexts --output='name')"
  if [ -f "$KUBECTX" ]; then
      # show '-' only if there's a saved previous context
      local PREV=$(cat "${KUBECTX}")

      _arguments \
        "-d:*: :(${all_contexts})" \
        "(- *): :(- ${all_contexts})"
  else
      _arguments \
        "-d:*: :(${all_contexts})" \
        "(- *): :(${all_contexts})"
  fi
}
```

**kubens.zsh**
```
#compdef kubens kns=kubens

__kubens() {
  _arguments "1: :(- $(kubectl get namespaces -o=jsonpath='{range .items[*].metadata.name}{@}{"\n"}{end}'))"
}
```

ahmetb:
It doesn't a throw an error for me on my oh-my-zsh setup with zsh 5.8 (x86_64-apple-darwin20.0).
Can we first try to figure out what causes the error?

martinmaehlmann:
The error is thrown, when the autocompletions are loaded, e.g. when I start a new instance of my zsh (zsh 5.7.1 (x86_64-apple-darwin19.0)).
Despite the error thrown, the completion itself does work.

The error thrown due to the fact that the `_arguments` are called outside a function. This is why I wrapped the completion scripts with an the functions `__kubens()` and `__kubectx()` silencing the error.

thatmarkenglishguy:
I was seeing this issue until I realised that rather than sourcing the completions they needed to be put into a directory on zsh's `fpath`
E.g. something like (untested)
```shell
# Create a folder ~/.zsh to put files zsh will load
mkdir "${HOME}/.zsh"
flamegraph --completions zsh > "${HOME}/.zsh/.flamegraph-completion.zsh"

# Add .zsh to your fpath when the zsh shell starts
echo 'fpath+=~/.zsh' >>"${HOME}/.zshrc"
```

AlmogBaku:
[...]

PR #377, opened 2022-12-16
Ignore errors of loading before completions initialized

AlmogBaku:
This PR fixes the issue I encountered on #285

It simply ignores the error of loading this script before the `compinit`.


ahmetb:
Is this pretty standard in comp scripts in zsh?

AlmogBaku:
I dunno, I saw it on another project that had the same issue...

ahmetb:
It would be good to research as it actually works fine for most people. Adding something we don't understand the consequences fully is going to create a problem in the future, especially if this bug happens because how things are set up for a small amount of people.

PR #381, opened 2023-02-24
Windows installation instructions (winget) (#365)

### LINKED ISSUE #365: Support winget
Right now, kubectx can only be installed on windows using Choco, however Windows also has its own package manager now called winget. It would be great if that was also supported.

Okeanos:
Resolves #365


ahmetb:
Thanks. Should I get added as a backup maintainer there?

PR #390, opened 2023-07-05
feat(tmux): add support for fzf-tmux in interactive mode

xero:
add support for tmux users via fzf-tmux


ahmetb:
Thanks for the submission. However, the shell script implementations are largely in maintenance mode as we're switching to Go programs. I recommend maintaining a fork for the time being if you plan on continuing to use these.

I suspect the right solution here is to offer an env var like KUBECTX_PICKER so the power users can pick between fzf, sk etc. #305

PR #395, opened 2023-07-19
Add manual installation instructions for latest Go releases to README.md

sylvainfaivre:
The Manual Installation for macOS and Linux contained only the older bash scripts, so add instructions for the Go releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,25 @@ winget install --id ahmetb.kubens
 
 ### Manual Installation (macOS and Linux)
 
-Since `kubectx` and `kubens` are written in Bash, you should be able to install
-them to any POSIX environment that has Bash installed.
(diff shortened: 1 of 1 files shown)


sylvainfaivre:
Well this MR's purpose was to give up to date instructions for the Go version.

When I installed this tool, I blindly followed the instructions from the README and installed the bash scripts.

Then a few months later, I wanted to see if there was an updated release, so I read the README again and was confused when I saw that I had in fact installed a much older release.

I feel that this MR fixes the instructions by providing an install procedure for the current version and making it clear that the bash scripts are the older version, and adding information about the completion scripts beeing needed in addition to the go binaries.

ahmetb:
Readme changed significantly since then. Closing.

PR #406, opened 2023-11-09
support delete context interactively

cvvz:
support delete context interactively.

Support using `kubectl ctx -d` to interactively choose a context to delete.


ahmetb:
this is a wontfix sorry, we don't need interactive mode on this tool (+bash implementations of kubectx/kubens are frozen at this point)

PR #409, opened 2023-11-24
Delete context with cluster and user

zcahana:
Adds an option to delete contexts including their referenced cluster and user entries from the `clusters/`users` lists.
The cluster and user entries are only deleted if they are no longer referenced in any other context (i.e., "orphaned").

The new option is exposed via a new CLI flag, `-D`, alongside the original flag (`-d`) which retains its existing functionality of deleting the context only.

### Implementation notes
- The toggle for the new behavior is captured by a boolean, termed "Cascade", which is propagated down from the CLI parsing logic, through the DeleteOp commands, and down to the `deleteContext` function.
- The bulk of the new logic is within the the `internal/kubeconfig` package, with new function for inspecting/manipulating `users`, `clusters` and their corresponding references within a `context`.
- The implementation for clusters and users is virtually similar except for the field names. In principal there's room for code sharing between these two, but given the scope of the project I figured we're better off with simple and straight-forward coding despite some code duplications.
- The new command is only available in the Go-based tool.

### Testing
- Added unit tests for all new functions in the `internal/kubeconfig` package.
- Added E2E tests via bats. Note that these should pass with the Go-based tool, but fail with the Bash tool. I'm not sure whether there's existing automation that runs these tests and with which of the kubectx implementations. If this is a problem, I'll come back to this to see what can be done.


zcahana:
@ahmetb hi, did you happen to have a chance to look at this? I'm wondering if this is something that may be of interest for this project.

PR #422, opened 2024-06-13
rename internal pkg

ReallyLiri:


ahmetb:
`internal` is a magic keyword in Go. :)

PR #416, opened 2024-01-21
feat(kubens): added force flag to switch namespaces even if it doesn'…

idsulik:
Hello!
Added `--force` && `-f` flags to `kubens` command to switch namespace even if it doesn't exist.
issue: https://github.com/ahmetb/kubectx/pull/333

### usage example
```
$ kubens not-found-namespace --force
Context "test" set.
Active namespace is "not-found-namespace".
---
$ kubens not-found-namespace --f
Context "test" set.
Active namespace is "not-found-namespace".
```

PR #425, opened 2024-07-10
chore: upgrade Go version to 1.22

idsulik:
In this PR https://github.com/ahmetb/kubectx/pull/416 I added `slices.Contains`, but it's not available for Go 1.20, replaced it with custom func and added `todo`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Setup Go
       uses: actions/setup-go@v2
       with:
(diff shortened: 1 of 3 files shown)


ahmetb (commented):
Can we upgrade Go version instead?
I suspect goreleaser file and github action and go.mod has it.

idsulik:
@ahmetb done

ahmetb:
The action needs to be updated too, it seems to be failing.

idsulik:
updated it as well

PR #429, opened 2024-09-22
Feat: add a flag to allow prefixing a query in interactive mode

Itaykal:
### Behaviour 
```sh
$ kubectx -q prod
```
will output: 
Screenshot 2024-09-22 at 18 16 59

### DIFF
--- a/cmd/kubectx/flags.go
+++ b/cmd/kubectx/flags.go
@@ -45,7 +45,7 @@ func parseArgs(argv []string) Op {
 			if len(argv) > 1 {
 				return InteractiveSwitchOp{SelfCmd: os.Args[0], Query: argv[1]}
 			} else {
-				return UnsupportedOp{Err: fmt.Errorf("'-q' needs arguments")}
+				return UnsupportedOp{Err: fmt.Errorf("'-q' needs an argument")}
 			}
 		} else {
 			return UnsupportedOp{Err: fmt.Errorf("'-q' only works in interactive mode")}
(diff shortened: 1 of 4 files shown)


ahmetb:
I'm failing to understand the use case for this. I never needed something like this and fzf does a good job of filtering the same results without introducing a flag with the same amount of keystrokes, no?

Itaykal:
> I'm failing to understand the use case for this. I never needed something like this and fzf does a good job of filtering the same results without introducing a flag with the same amount of keystrokes, no?

Well yes but actually no, I want to be able to create the following aliases:
```
kx=kubectx -q <staging-context-identifier>
kxp=kubectx -q <production-context-identifier>
```
And create some "segregation" between context "types", in order to avoid logging in to the wrong cluster

ahmetb:
I think you're trying to solve the problem at the wrong layer. I recommend segregating by different kubeconfig files entirely. 

Also, this sort of problem won't solve accidents you're trying to prevent. Speaking from experience, a writable context entry is always a recipe for disaster. 

At any rate, I don't think this concern should be addressed at this tool, and if you use separate kubeconfig files it would work with this tool with no modifications.

zcahana:
> Well yes but actually no, I want to be able to create the following aliases
```
kx=kubectx -q <staging-context-identifier>
kxp=kubectx -q <production-context-identifier>
```

@Itaykal if your use case is such aliases, you could achieve something similar via:
```
kx='FZF_DEFAULT_OPTS="-q <staging-context-identifier>" kubectx'
kxp='FZF_DEFAULT_OPTS="-q <production-context-identifier>" kubectx'
```

PR #432, opened 2024-11-25
Add filtering contexts

malarzm:
In our work we're using separate contexts for each shop we're running. We have a custom tool that automates running same commands across multiple apps, it boils down to being a fancy `for` loop over `kubectx`'s contexts :) Lately we've found ourselves in a situation we need to be able to run things across all shops but only within one cluster. To keep shop vs cluster configuration in one place we thought it'd be nice if `kubectx` could filter contexts for us, thus this PR.

Disclaimer: this is very first time I've written in Go so I could have misplaced/misnamed/missed some obvious stuff. I'll be updating this PR with all suggestions you may have


internal/kubeconfig/contexts.go
```diff
@@ -44,7 +44,7 @@ func (k *Kubeconfig) contextNode(name string) (*yaml.Node, error) {
 	return nil, errors.Errorf("context with name \"%s\" not found", name)
 }
 
-func (k *Kubeconfig) ContextNames() []string {
+func (k *Kubeconfig) ContextNames(filters map[string]string) []string {
```
malarzm, line 47: I'm open to suggestions how you want to handle optional parameters, be it a struct or separate function, or anything else

internal/kubeconfig/contexts.go
```diff
@@ -54,8 +54,19 @@ func (k *Kubeconfig) ContextNames() []string {
 	}
 
 	var ctxNames []string
+ctxLoop:
```
malarzm, line 57: I believe this could be refactored to avoid syntax that feels like a ` goto` to me, maybe having the filtring as a separate function. I'm open to suggestions

malarzm:
@ahmetb is this functionality something you'd even consider merging? We'll work on something on our end if it's not a good fit

ahmetb:
Have you tried the fzf integration in kubectx? I think it does a lot of what you're trying to achieve here without us having to maintain extra code.

malarzm:
If I understand correctly, `fzf` allows using interactive mode. In our case there will be no interactivity, as it's a script that needs to get filtered contexts. To illustrate, we'd like to have `kubemob cluster=c1 disable-cron xxx` (`kubemob` being a fancy `for` loop I've mentioned in the original description) and it should

1. Somehow get all contexts that operate within `cluster: c1`
2. Do its thing (disable cron) for all contexts

I'm not sure how to utilize `fzf` in this scenario

ahmetb:
Sorry but kubectx isn't meant to be scripted around. This is explained in the README. :) We dont provide a guaranteed output format. Use kubectl directly for that.

malarzm:
All right, thank you for your time!

PR #461, opened 2026-03-08
feat: add -s/--shell flag for scoped sub-shell

### LINKED ISSUE #12: eval-able and idempotent kubectx
What do people following `kubectx` think of this idea:

We can detect if `kubectx` command is executed in interactive mode OR eval'ed in a shell (like `eval $(kubectx ...)`, piped elsewhere or redirected to a file, by checking if /dev/stdout is a TTY or not.

**If kubectx is evaluated in a shell, we can print an `alias` that aliases kubectl to `kubectl --context=NAME`. So when that shell is exited, the default context would go back to what's set before. This would let users work on multiple clusters simultaneously in different terminal tabs. It would look like this:**

    eval $(kubectx NAME)

and it would print something like `alias kubectl=kubectl --context=NAME` which then is evaluated in the current shell, therefore creating an alias. Similarly, `eval $(kubens ...)` would also evaluate an `alias kubectl=kubectl --namespace=NAME` (these two wouldn't work together, but I might have a solution for that too, by writing a basic bash function that picks up these names from env.

Just an idea. The good old way of using kubectx/kubens would still be around and would work the same way.

ahmetb:
Spawns an isolated sub-shell with a minimal kubeconfig containing only the specified context.

This allows the user to launch a shell where they can only interact with a single cluster without having to worry about a command or an LLM agent interacting with other contexts.

Inside the isolated shell, most context switching/editing operations on kubectx are blocked. Nested shells not allowed.

Fixes #12.


bmjask:
👍

